PR #34, opened 2021-08-10
Npm

hanayik:
update version, update robust range, disable alpha blend (can put back in if needed).


cdrake (approved):
I had a question about disabling blend.  We can chat about it later.

PR #40, opened 2021-08-12
3dfix

hanayik:


cdrake (approved):
We should also address RGB values for min max.

hanayik:
Sounds good! Could you please add RGB min max to the issue board!

PR #47, opened 2021-08-13
Sync

hanayik:
- add `syncWith` method. Allows syncing crosshair location, azimuth, and elevation between multiple niivue canvases
- add syncWith test
- removed snapshot from test that didn't need it


hanayik:
I need to update the github action recipe so that the artifacts are uploaded even if the test fails. This is needed so that I can see the failed diffs from screenshots. (that's why this PR fails some checks). It did the right thing at least! Just need to see how off it is. I suspect the mouse interaction could be an issue....

hanayik:
Should be good to go now. I was having issues because the whole page screenshot in the CI environement was no the same size as the page screenshot in the local tests. To avoid this, I only take a screenshot of the canvas element that I expect to have changed because it is synced with another canvas element. Therefore, we are now only comparing the canvas that we expect to update as a result of clicking in a completely different canvas.

PR #60, opened 2021-08-17
Render multiple objects

cdrake:
This adds a for loop in the draw3D function and niivue object and shader classes to capture render options.  It also computes the MVP matrix for all geometries added as objects.


hanayik (approved):
Looks good. Bumping the version up as well. I've been updating the minor number when there's rather large changes, and the patch (major.minor.patch format) for small changes until we reach a stable API at version 1. If you ever bump the version in the future, just keep the major number at 0 until everything is more stable in a few months.

PR #74, opened 2021-08-20
update docs and demos

hanayik:


hanayik:
I have added a new test as part of this commit as well. the new test adds the hippo overlay on top of the mni background image so that we can avoid this sort of issue in the future (it has generated a new snapshot). Hopefully any overlay issues will be caught sooner next time.

PR #81, opened 2021-08-31
Feature object picking

cdrake:


hanayik (commented):
@cdrake , Just wanted to make sure I understood the new feature here. In the browser console, I get an id of "1" when clicking on the 3D clip plane, and and id of "2" when clicking on the brain (or just outside the brain). This way, we can now differentiate between clip plane clicks (and touches) and brain clicks right? and therefore, add in clip plane click and drags later right? Just checking!

hanayik (approved):

PR #84, opened 2021-09-07
update color map formats

hanayik:
luts are now defined in JSON files using a strict set of keys. This makes querying available color maps easier via the new `niivue.colorMaps()` function. I have also added a new demo to show all color maps. To add new color maps in the future the process is:

- add a new JSON file to `src/cmaps/`
- import that JSON object into `src/cmaps/index.js` just like the other maps that exist

related to #83


hanayik:
@cdrake , I have updated this PR by rebasing to include your latest object picking changes. Needed to be done since we had differing versions of the main branch.

PR #100, opened 2021-09-20
Feature depth picking

cdrake:


hanayik (changes requested):
@cdrake, there seems to be a major performance issue in this PR. 

I checked out this branch locally and played around. The 3D demos always crashed firefox on my mac, and in chrome the webgl2 context was lost (but didn't crash). Either way, both tests point to a 3D rendering issue somewhere.

I'm not sure where it's coming from (maybe the shader, or the changes in draw3D?). The issue is only present in 3D views, so there's some evidence to point to 3D drawing routines. To elaborate, the performance issue only occurs when clicking and dragging the 3D rendering in order to rotate it (in any direction). In my experience testing this just now, the first few seconds seem to go ok (I can rotate the image some) but very quickly the drawing stutters and then freezes the entire OS until the browser figures out what it wants to do (crash, or lose the context). This happens in any 3D demo. 

Do you mind investigating where the issue is coming from? Did you experience any issues like this locally on your machine (I know you're on Ubuntu right)?

cdrake:
I’ll check it out. I did not see these issues locally on my Ubuntu machine.

hanayik:
@cdrake, It seems that this is only an issue on one machine I have. I just tested again on various devices and here are the results (check means it works without issue)

- [x] macOS Big Sur Chome
- [x] macOS Big Sur Firefox
- [ ] macOS Big Sur Safari 14 (crosshairs do not update in sync.3dplanar.html)
- [x] Android Chrome

on macOS Big Sur (Safari 14) the rendering is fine, but the crosshairs in multiplanar sync do not update. Also, I think the object picking is not working in that macOS + Safari env because the console reports `object id is 0` in Safari 14, but on the same machine in firefox the `object id is 2` for the brain. 

So, it looks like things are mostly working on a variety of devices. However, my macOS catalina machine (in any browser) has major issues, and that was where I did the first round of testing earlier this morning. I'm not sure this is something we can directly address, but it might be nice to try and replicate. 

@neurolabusc , do you have access to an older (non M1) mac, that is macOS < Big Sur to test on?

neurolabusc:
@cdrake there are iMacs in the BatCave you can use. There is a modern Intel iMac with Big Sur, as well as a couple older Macs that can not be upgrade to modern macOS as their GPUs do not support Metal. The only caveat is that the HVAC is out in the bat cave at the moment, I can either ping you when temperature is restored or you can move the computers to the ABC suite.

cdrake:
@neurolabusc, thank you. Those machines will provide a good test bed.

neurolabusc:
The Intel-based MacOS is a strange edge case. I believe that MacOS WebGL is built on their Metal API. The Metal API is really bespoke to Apple's own in house GPUs, found in the iPhone A-series and Mac M-series SOCs. The API was back ported to some Intel-based computers, but not all Metal features are [supported](https://developer.apple.com/documentation/metal/rendering_a_scene_with_deferred_lighting_in_objective-c). As an aside, Apple's GPUs do not support all OpenGL features, which explains why M1 Macs can not support [geometry shaders](https://github.com/MRtrix3/mrtrix3/issues/2247).

hanayik (approved):
Hi @cdrake, 

Looks good. I'm not sure why the github test runner is hanging, but I have tested locally so I'm fairly confident that we're good. Also, just a note, the new depth picking seems to only work with Safari 15+. I tested in safari 14 and 15, and only 15 worked well. Might be good to update the readme with that info saying that Safari 15 is the minimum we support going forward. Anyone with the latest macOS should have it.

PR #171, opened 2021-11-20
Packaging

hanayik:
List of changes:

- add "v" hot key for changing view modes
- add new default option for keyup debounce time (50 ms) (useful fro "c", and "v" hot keys)
- update module bundler to vite
- add `npm run dev` script for blazing fast hot module reload during development (changes to src update automatically in the browser when using npm run dev)
- remove old font resources
- update tests and demos to use built umd bundle
- no longer distribute niivue as a prebuilt bundle 
- bundling is now the responsibility of the downstream devs (like most packages)
- vite, and Webpack work well (Webpack tested with vue-cli)
- import fonts as resources rather than hard coded paths (good for bundlers)
- some minor code edits related to object and array destructuring
- externalised dependencies
- removed all console.log statements (or commented out)
- update version


src/niivue.js
```diff
@@ -158,8 +158,8 @@ export const Niivue = function (opts = {}) {
   this.otherNV = null; // another niivue instance that we wish to sync postion with
   this.volumeObject3D = null;
   this.clipPlaneObject3D = null;
-  this.crosshairPosition$ = new Subject();
-  this.intensityRange$ = new Subject();
+  this.intensityRange$ = new Subject(); // needs to be updated to have an intensity range for each loaded image #172
+  this.scene.location = new Subject(); // object with properties: {mm: [N N N], vox: [N N N], frac: [N N N]}
```
cdrake, line 162: I was wondering why this is not named this.scene.location$.

cdrake (approved):

src/niivue.js
hanayik: Hi @cdrake , The $ convention is new to me. But since it's the known convention in rxjs I will add it back it. It does visually help identify that the property is an observable, so it's a good idea to use the $. I'll update and merge in another PR soon.

PR #204, opened 2021-12-12
Feature c wasm

cdrake:


src/niivue.js
```diff
@@ -1354,6 +1356,125 @@ Niivue.prototype.initText = async function () {
...
+  } else {
+    this.refreshLayers(processedImage, imageIndex, this.volumes.length);
+  }
+  this.drawScene();
+
+  this.updateGLVolume();
+};
+
+Niivue.prototype.initWasm = async function () {
+  this.linearMemory = new LinearMemory({ initial: 2048, maximum: 2048 });
```
neurolabusc, line 1470: I would suggest a smaller initial of 256, since we can grow memory.

cdrake: Updated

hanayik:
@cdrake , and @neurolabusc, the niimath wasm is looking good. However, niivue is still not working in Safari 15.1 (the default on any new MacBook). And it does not work on Safari in iOS 15.1. 

Before wasm, all of Niivue worked in all major browsers, so we should investigate this before potentially merging a breaking change.

src/niivue.js
hanayik: Although the initial memory was changed here, was the WASM recompiled using Chris R's latest compile command?

demos/features/niimath.command.html
```diff
@@ -1,120 +1,267 @@
...
-    <script src="../niivue.umd.js">
-    </script>
-    <script>
+      <button class="process-button" onclick="process()">Process</button>
+    </div>
+  </section>
+  <section style="margin-top: 120px;">
+    <pre>
+        Datatype information:
+ -dt sets the datatype used internally for calculations (default float for all except double images)
```
neurolabusc, line 98: dt and dot does not apply to WASM

demos/features/niimath.command.html
```diff
@@ -1,120 +1,267 @@
...
+ -tensor_2upper           : convert NIfTI standard lower triangle image to FSL style upper triangle order
+ -tensor_decomp_lower     : as tensor_decomp except input stores lower diagonal (AFNI, ANTS, Camino convention)
+ -trunc                   : truncates the decimal value from floating point value and returns integer value
+ -unsharp  <sigma> <scl>  : edge enhancing unsharp mask (sigma in mm, not voxels; 1.0 is typical for amount (scl))
+ -dog <sPos> <sNeg>       : difference of gaussian with zero-crossing edges (positive and negative sigma mm)
+ -dogr <sPos> <sNeg>      : as dog, without zero-crossing (raw rather than binarized data)
+ -dogx <sPos> <sNeg>      : as dog, zero-crossing for 2D sagittal slices
+ -dogy <sPos> <sNeg>      : as dog, zero-crossing for 2D coronal slices
+ -dogz <sPos> <sNeg>      : as dog, zero-crossing for 2D axial slices
+ --compare <ref>          : report if images are identical, terminates without saving new image
```
neurolabusc, line 128: compare does not apply to WASM

demos/features/niimath.command.html
```diff
@@ -1,120 +1,267 @@
...
+ -ero     : Erode by zeroing non-zero voxels when zero voxels found in kernel
+ -eroF    : Minimum filtering of all voxels
+ -fmedian : Median Filtering 
+ -fmean   : Mean filtering, kernel weighted (conventionally used with gauss kernel)
+ -fmeanu  : Mean filtering, kernel weighted, un-normalized (gives edge effects)
+ -s <sigma> : create a gauss kernel of sigma mm and perform mean filtering
+ -subsamp2  : downsamples image by a factor of 2 (keeping new voxels centered on old)
+ -subsamp2offc  : downsamples image by a factor of 2 (non-centered)
+
+Dimensionality reduction operations:
```
neurolabusc, line 211: Dimensionalty reduction, subsamp, etc. do not exist in WASM. Only commands listed as `YES` in [this table](https://github.com/rordenlab/niimath/blob/development/src/JavaScript.md) are supported.

Issue #163, opened 2021-10-31
3D render mode stalls github actions tests


hanayik:
through some iterative debugging, I have determined that the automated github actions tests are broken for all tests that set the `sliceType` to `sliceTypeRender`. This means all 3D rendering tests stall (not fail) the github actions runner indefinitely. 

@cdrake, can you think of any changes a couple of months ago to the render loop that might have caused this? I have not found the cause of this "test stalling" behavior yet, just narrowed it down to the 3D render cases. Maybe we can brainstorm during our next meeting about some possible places to check in the code. 

All tests do still run locally on any computer when using `npm run test`. So that's not an issue. It would be good to try and solve why 3D renderings are failing in the CI environment though.

cdrake:
Sounds good.  It seems difficult to diagnose since it passes locally.  Happy to brainstorm.

PR #245, opened 2022-02-21
change runner to macos

hanayik:


hanayik:
this PR resolves #163. The runner is now self hosted on a macOS machine

PR #248, opened 2022-02-21
Robust Affine (NIfTI methods 1,2,3: pixdim, qform, sform)

neurolabusc:
The NIfTI header provides [3 methods for defining a spatial transform](https://nifti.nimh.nih.gov/pub/dist/src/niftilib/nifti1.h). This patch will choose between the SForm or the QForm depending on which has a higher xform code. In the case of a tie, precedence is given to the SForm unless the flag `useQFormNotSForm` is set to true (in which case the QForm is used regardless of xform codes). If this transform does not make sense, the software will fall back to the Analyze style of using only the pixdim for spatial information (this is unlikely to be a good result).


neurolabusc:
New option `isNearestInterpolation` allows images to be shown with nearest neighbor or linear interpolation. This is experimental: we might want some discussion regarding whether atlases are always nearest neighbor; whether interpolation is only for 2D slices or also for 3D rendering, etc.

I also removed the `Contributors` from the README, as these are explicitly listed on the right column.

Issue #213, opened 2022-01-26
implement label image region outlining similar to FSLeyes


hanayik:
will be similar to this shader 
https://github.com/rordenlab/rordenlab.github.io/blob/13cd4544339ee898f7cbf531824b79068a08df3b/volume/shader-srcs.js#L304

PR #250, opened 2022-02-23
Experimental Atlas support

neurolabusc:
- Automatically detects Atlases where the [NIfTI header intent](https://nifti.nimh.nih.gov/pub/dist/src/niftilib/nifti1.h) is `NIFTI_INTENT_LABEL ` (1002). For example, the atlaes supplied with MRIcroGL
 - Option isAtlasOutline determines if atlas is drawn as an outline ([issue 213](https://github.com/niivue/niivue/issues/213)).
 - Similar to MRIcroGL, label colors repeat after 100 (e.g. same colors repeat for indices 1..100, 101..200, 201..300). The color should really be set to `Random`. @hanayik if someone drags-and-drops a NIfTI image onto NiiVue, it starts with the default color scheme `gray`: can you change this to default to `random` if the intent code is 1002?
 - We still need to know how to include the label names, e.g. 1Left Inferior Parietal Lobe`. MRIcroGL includes this text in the NIfTI header, between the 352 byte binary header and the image data. Maybe @cdrake's work on the AFNI extensions can be extended to support this (MRIcroGL requires three features to detect this: the `regular` character is 98 (not the typical 114), the intent is labels, and the vox_offset is greater than 352.
 - At the moment only UINT8 images are supported, easy to extend to other integer data types (e.g. INIA19 Primate Brain Atlas).

VX2

### DIFF
--- a/src/niivue.js
+++ b/src/niivue.js
@@ -13,6 +13,7 @@ import {
   fragOrientShaderI,
   fragOrientShaderF,
   fragOrientShader,
+  fragOrientShaderAtlas,
   fragRGBOrientShader,
   vertSurfaceShader,
   fragSurfaceShader,
@@ -74,6 +75,7 @@ export const Niivue = function (options = {}) {
     viewModeHotKey: "KeyV", // keyboard shortcut to switch view modes
     keyDebounceTime: 50, // default debounce time used in keyup listeners
     isNearestInterpolation: false,
+    isAtlasOutline: false,
     isRadiologicalConvention: false,
     logging: false,
   };
@@ -89,6 +91,7 @@ export const Niivue = function (options = {}) {
   this.colorbarShader = null;
   this.fontShader = null;
   this.passThroughShader = null;
+  this.orientShaderAtlasU = null;
   this.orientShaderU = null;
   this.orientShaderI = null;
   this.orientShaderF = null;
@@ -1544,6 +1547,12 @@ Niivue.prototype.init = async function () {
     fragPassThroughShader
   );
 
+  this.orientShaderAtlasU = new Shader(
+    this.gl,
+    vertOrientShader,
+    fragOrientShaderU.concat(fragOrientShaderAtlas)
+  );
+
   this.orientShaderU = new Shader(
     this.gl,
     vertOrientShader,
@@ -1751,6 +1760,8 @@ Niivue.prototype.refreshLayers = function (overlayItem, layer, numLayers) {
   let orientShader = this.orientShaderU;
   if (hdr.datatypeCode === 2) {
     // raw input data
+    if (hdr.intent_code === 1002)
+      orientShader = this.orientShaderAtlasU;
     this.gl.texStorage3D(
       this.gl.TEXTURE_3D,
       1,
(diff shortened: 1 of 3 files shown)


hanayik:
This PR has failed the automated test suite. Specifically the RGB image test. 

The diff snapshot from the test is attached below (original on the left, diff in the middle, and newly generated on the right)

there is now a L/R flip. Could this be due to the changes in the orientation shader in this PR?

test-read-rgb-multiplanar-js-read-rgb-multiplanar-1-diff

hanayik:
> * Automatically detects Atlases where the [NIfTI header intent](https://nifti.nimh.nih.gov/pub/dist/src/niftilib/nifti1.h) is `NIFTI_INTENT_LABEL ` (1002). For example, the atlaes supplied with MRIcroGL
> * Option isAtlasOutline determines if atlas is drawn as an outline ([issue 213](https://github.com/niivue/niivue/issues/213)).
> * Similar to MRIcroGL, label colors repeat after 100 (e.g. same colors repeat for indices 1..100, 101..200, 201..300). The color should really be set to `Random`. @hanayik if someone drags-and-drops a NIfTI image onto NiiVue, it starts with the default color scheme `gray`: can you change this to default to `random` if the intent code is 1002?
> * We still need to know how to include the label names, e.g. 1Left Inferior Parietal Lobe`. MRIcroGL includes this text in the NIfTI header, between the 352 byte binary header and the image data. Maybe @cdrake's work on the AFNI extensions can be extended to support this (MRIcroGL requires three features to detect this: the `regular` character is 98 (not the typical 114), the intent is labels, and the vox_offset is greater than 352.
> * At the moment only UINT8 images are supported, easy to extend to other integer data types (e.g. INIA19 Primate Brain Atlas).
> 
> VX2

@neurolabusc, yes I can set the colormap to random based on the intent code. I will make the change today in a separate PR

neurolabusc:
@hanayik this is patched for now, I simply reverted the application of the transformation matrix for RGBA images. This is not a long term solution, we need more RGBA rotation examples. FSLeyes and MRIcroGL both apply spatial transforms to RGB images, but they appear to have a LR flip relative to NiiVue. 

The relevant line is the shader, where one expects the vertex location to be transformed with the spatial transformation matrix `mtx`:

```
vec4 vx = vec4(TexCoord.xy, coordZ, 1.0);// * mtx;
```

neurolabusc:
@hanayik the shader **should** apply the transform
```
vec4 vx = vec4(TexCoord.xy, coordZ, 1.0) * mtx;
```
and the results are consistent with MRIcroGL and FSLeyes. The text 'MTT' in the image should look normal when viewed in radiological convention and mirror-reversed when viewed with neurological convention. I had just forgot the default for the `npm run dev` is neurological.

So I would suggest:
 1. Either have the test ensure the image is created with radiological convention or replace the test.
 2. Can I suggest you run `pigz -n -11 ct_perfusion.nii` to compress this test to nii.gz - this reduces the file size dramatically.

hanayik:
@neurolabusc, good catch. I will update the test in this case and compress the test image.

PR #254, opened 2022-03-01
Feature 3d crosshairs

cdrake:


hanayik:
The user should be able to pass an option to the Niivue constructor to show or hide the 3D crosshair. I'm adding this in now. I think the default should be `show3Dcrosshair: false`.

neurolabusc (commented):
@hanayik I think the 3D crosshairs should also be enabled in the sync.3dplanar.html demo.

hanayik:
@neurolabusc , I updated the demo already :)

PR #264, opened 2022-03-07
Update verison and actions

hanayik:

- add windows GitHub actions test
- runs windows test in headless mode

Demos work on:

- [x] iOS Safari
- [x] iOS Firefox
- [x] iOS Chrome
- [x] Android Chrome
- [x] Android Firefox
- [x] macOS Safari
- [x] macOS Firefox
- [x] macOS Chrome


hanayik:
@neurolabusc, I noticed that when you zoom in with the 3D view active, the crosshair location changes based on zoom level.  If you zoom to the max level, set the crosshair, and then zoom out to the min level, the crosshair will be rendered in a new location

PR #286, opened 2022-03-18
Single vertex buffer and VAO for volume rendering, load connectome fr…

neurolabusc:
Single vertex buffer and VAO for volume rendering, load connectome from JSON

### DIFF
--- a/src/nvmesh.js
+++ b/src/nvmesh.js
@@ -4,6 +4,7 @@ import { v4 as uuidv4 } from "uuid";
 import * as cmaps from "./cmaps";
 import { Log } from "./logger";
 import * as mat from "gl-matrix";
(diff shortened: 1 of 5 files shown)


src/nvmesh.js
```diff
@@ -3,8 +3,6 @@ import * as pako from "pako";
 import { v4 as uuidv4 } from "uuid";
 import * as cmaps from "./cmaps";
 import { Log } from "./logger";
-import * as mat from "gl-matrix";
-import { NiivueObject3D } from "./niivue-object3D.js";
```
neurolabusc: NiivueObject3D is a required dependency: used by loadConnectome to generate cylinder and sphere geomtery.

PR #289, opened 2022-03-25
Support NRRD/NHDR, MGH/MGZ, AFNI HEAD/BRIK voxel formats. STL, legacy…

neurolabusc:
Support NRRD/NHDR, MGH/MGZ, AFNI HEAD/BRIK voxel formats. STL, legacy VTK mesh formats.

Note that for paired images (NHDR, HEAD/BRIK) the user must supply the header as the `url` (header file) and the  `urlImgData` (image file).


hanayik:
@neurolabusc, the orientation labels are not correct when reading MGH. The same file opened in MRIcroGL or FSLeyes shows correct orientation labels. 

Screenshot 2022-03-28 at 12 02 17 pm

[wm.mgz.zip](https://github.com/niivue/niivue/files/8362066/wm.mgz.zip)

neurolabusc:
@hanayik - well spotted. This should be fixed with [the latest commit](https://github.com/neurolabusc/niivue/commit/86a86345129f193381a59b6b25be7f12a839f4c2). The documentation for this format is sparse, and it is unclear if the origin is the first voxel's corner or center. It might be worth checking that NiiVue generates the same SForm solution as FSLeyes. 

Since FSL and [FreeSurfer](https://github.com/freesurfer/freesurfer/blob/dev/LICENSE.txt) do not use permissive licenses, I am unable to validate my formula without compromising our license. I strongly believe that research is based on reproducibility, and I do hope we can move beyond these black boxes. I wish FreeSurfer was `free` as in freedom, not free as in dollars. Hopefully, modern NIH resource sharing plans will prevent these situations in the future. I note that unlike most of FSL, FSLeyes does use a truly free license. However, looking at the code in Github it looks like it relies on other tools for this formula.

Perhaps @dngreve can document the transform. 

You could also check that my solution matches that of FreeSurfer's conversion:

```
mri_convert wm.mgz wm.nii.gz
```

neurolabusc:
@[dngreve](https://github.com/dngreve) thanks for your rapid response. My ambiguity in converting MGH spatial transforms to NIfTI [is this line](https://github.com/niivue/niivue/blob/e67273337430a378a41d6753d91364e9e89b4d33/src/nvimage.js#L453) which influences the translation of the image. I do try to be very cautious of software contamination.

neurolabusc:
OK, that makes sense, so the current code is correct (and I did test that it seems to match FSLeyes).

Issue #287, opened 2022-03-21
Clip plane face seems to have some blending that can look distorted


hanayik:
<img width="943" alt="Screenshot 2022-03-21 at 1 52 31 pm" src="https://user-images.githubusercontent.com/5217720/159275709-a9362752-43a2-4d07-a6b2-481f0d69ada9.png">
[T1.nii.gz](https://github.com/niivue/niivue/files/8315934/T1.nii.gz)

PR #293, opened 2022-03-28
Basic tractography

neurolabusc:
- Adds support for tractography streamlines (TRK/TCK). The API still needs work (they are now `meshes`, but have unique properties and shaders). However, the basic implementation is solid. In general, there are four ways to represent streamlines with a graphics API, listed in ascending demand on resources:

 - [Line strip](https://www.labri.fr/perso/nrougier/python-opengl/#gl-primitives), makes minimal demands on GPU, but lines limited to one pixel in thickness.
 - [bill board imposters](https://www.springerprofessional.de/en/a-simple-and-efficient-cylinder-imposter-approach-to-visualize-d/14979844) requires about three times the resources of line strips but allow thick lines with emulated lines.
 - geometry shaders like MRtrix dixel views. Geometry shaders do not exist in WebGL (or Metal).
 - Triangulated mesh cylinders - this would use the same shader as all NiiVue meshes, but places the highest demand on GPU.

The current implementation uses line strip with just 16 bytes per vertex (XYZ positions as 32-bit float and 32-bit RGBA color). In the future, we might want to add support for imposters, but this may not be suitable to phones and other portable devices.


neurolabusc:
Taylor,
  The rationale for the changes observed for the [tests](https://github.com/niivue/niivue/pull/293/commits/eeb381f88b2e02b34c22f66731694c2d0c0ea39c) is an experimental solution for [issue 287](https://github.com/niivue/niivue/issues/287). The clip plane is now slightly translucent over tissue. If we like this, we should think of an easy way to influence the visibility of clip planes (e.g. they vanish after a few seconds, or some gesture to toggle clip plane visibility).

Issue #344, opened 2022-06-16
3D rendering failed


BastienCagna:
Hi,
The 3D rendering of https://niivue.github.io/niivue/features/basic.3d.html failed on my computer with firefox and chrome:
![image](https://user-images.githubusercontent.com/22853356/174103242-df4e68fb-2fb2-4dce-ae78-6db1ad202671.png)

Here are the webglreport for WebGL 2 with firefox:
![image](https://user-images.githubusercontent.com/22853356/174102464-71fe00d3-86c2-4e0d-a7eb-2cb127b34958.png)

With chrome:
![image](https://user-images.githubusercontent.com/22853356/174102736-23a2c97b-e77c-4e5f-82bf-fd21968fe964.png)

Finally, my computer infos
![image](https://user-images.githubusercontent.com/22853356/174102993-dc283d3b-372f-4a4f-aa19-2e743ed5ffea.png)

alexisthual:
I have the exact same problem!
Here is my GPU info:
ANGLE (Intel Open Source Technology Center, Mesa DRI Intel(R) UHD Graphics 620 (Kabylake GT2) , OpenGL 4.5)

hanayik:
Hi @alexisthual ,

can you try running this in a terminal session:

```
MESA_GL_VERSION_OVERRIDE=3.3 firefox # or brave, chrome, etc.
```

then navigate to the niivue 3D demo. 

https://niivue.github.io/niivue/features/basic.3d.html

alexisthual:
> can you try running this in a terminal session: 
> ```
> MESA_GL_VERSION_OVERRIDE=3.3 firefox # or brave, chrome, etc.
> ```

I just did and unfortunately got the same rendering artefact as before!

alexisthual:
Actually, I tried different browsers: https://niivue.github.io/niivue/features/basic.3d.html displays correctly for me in firefox and chrome, but not in brave.

neurolabusc:
@cdrake can you replicate this on your system? I think we can fix this if we have access to a system that shows this bug. Reminds me of the [MRIcroGL kludge for Mesa](https://github.com/rordenlab/MRIcroGL/blob/4c21503f4fb5cd10ed59382ad8085084ea6a54f2/glvolume2.pas#L321) which adds a GLSL conditional that is never executed but make Mesa behave nicely.

hanayik:
[...]

PR #349, opened 2022-06-22
BF: Avoid use of continue statement in raycast loop, to work around apparent mesa bug

pauldmccarthy:
This PR fixes the artifact reported in #344. 

However, I don't know enough about the niivue internals to understand the ramifications of removing these clauses.  Specifically, I don't know where the `<voxel intensity>->RGBA` transfer function is defined, nor what the resulting alpha values represent.  These clauses were present in the very first commit on the niivue/niivue github repository, so there is no provenance as to what they intend to accomplish. @neurolabusc, do you have any insight here?

- #344 

Demos work on:

- [ ] iOS Safari
- [ ] iOS Firefox
- [ ] iOS Chrome
- [ ] Android Chrome
- [ ] Android Firefox
- [ ] macOS Safari
- [ ] macOS Firefox
- [ ] macOS Chrome
- [ ] Windows Chrome
- [ ] Windows

### DIFF
--- a/src/shader-srcs.js
+++ b/src/shader-srcs.js
@@ -130,12 +130,13 @@ void main() {
 	while (samplePos.a <= len) {
 		vec4 colorSample = texture(volume, samplePos.xyz);
 		samplePos += deltaDir; //advance ray position
-		if (colorSample.a < 0.01) continue;
-		if (firstHit.a > lenNoClip)
-			firstHit = samplePos;
-		backNearest = min(backNearest, samplePos.a);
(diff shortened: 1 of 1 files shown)


neurolabusc:
@pauldmccarthy nice detective work. I notice you have modified two of the three `continue` calls in the shaders. I suggest we fix all 3 to cope with mesa.

pauldmccarthy:
@neurolabusc Just pushed another commit, which adjust the two ray-cast loops in the `fragRenderShaderMIP` source - so that's four in total - does that cover everything?

PR #373, opened 2022-07-27
Allow mesh shader to be independently set for each mesh

neurolabusc:
- Change API so that different meshes can be assigned different shaders. As seen below (and in the new live demo), the cortex uses an outline shader to reveal the subcortical regions that are shown with a conventional shader.

shaders


demos/features/sync.mesh.html
```diff
@@ -41,28 +41,29 @@ <h1>
...
       }
       var nv1 = new niivue.Niivue({backColor: [1, 1, 1, 1]})
       nv1.attachTo('gl1')
       nv1.setHighResolutionCapable(false)
-      nv1.setMeshShader('Outline')
       nv1.opts.isOrientCube = true
-      nv1.loadMeshes([
+      await nv1.loadMeshes([
          {url: "../images/BrainMesh_ICBM152.lh.mz3"},
       ])
```
hanayik, line 55: We can prob do something like this:

```
nv1.loadMeshes([
  {url: "../images/BrainMesh_ICBM152.lh.mz3"},
]).
then((id)=>{
nv1.setMeshShader(id, 'Outline')
}
```

neurolabusc: @hanayik sounds fine to me. Do you want to implement this with the relevant /demos/features/*.html files.

PR #441, opened 2022-11-29
NVMesh losing passed name properties in loadFromUrl

bwagnerwork:

- loadFromUrl had name attribute overwritten causing URLs like "/data_cat12/lc.central.T1.gii?withJunk=true" to fail even if name was specified as "lc.central.T1.gii"


hanayik:
@bwagnerwork , thanks for pointing this out and suggesting a fix. We were dealing with query strings in NVImage but not in NVMesh. I have updated NVMesh to handle query strings better (along with query strings in mesh layers). Closing this PR and will submit a new PR with all changes.

Issue #482, opened 2023-01-19
Mesh rendering performance


neurolabusc:
@mrneont and @afni-dglen note that visualization is not fluid when Intel integrated graphics are used with [HighDPI displays](https://www.khronos.org/webgl/wiki/HandlingHighDPI).  This is seen in the [FreeSurfer demo with 164k vertices](https://niivue.github.io/niivue/features/mesh.freesurfer.html) which is sluggish on Intel Mac's with `retina` displays, but fluid on Apple Silicon Mac's with similar displays. Perhaps the `opts.isHighResolutionCapable` default should be initialized as `false` if the rendererInfo reports an Intel GPU.

A more complicated but sophisticated solution would be to use low resolution during dynamic motion and use high resolution when user interaction pauses (this is the default behavior of MRIcroGL).

mrneont:
@neurolabusc : that sounds good, thanks.

afni-dglen:
Thanks. Just to be sure, I'm seeing this seemingly erratic behavior on an Intel MacBook Air (2020). At higher resolution (4k or 1440p), it seems to be a little better than lower resolution (1080p). Most of this is from just too fast rotations. The rotation rate seems to increase as I move the mouse cursor toward the center and then slow down as I move away. It's more usable far from the center. 

On my M1 MacBook Air (2021), it works, but it's still a little faster than I would expect, but it's fluid. I'm using a 1000dpi mouse, and I think I get about 1-1/2 rotations moving the mouse very slowly from left to right across that FreeSurfer surface. If I move it quickly, I get many more rotations. I think in no case do we need a very fast rotation for this, so setting a limit on rotational speed seems like a good idea.

BTW, the FreeSurfer demo surface doesn't appear on my equally vintage Android phone.

neurolabusc:
Closing this issue. The latest release changes the defaults to trade-off quality for performance. The equipment of @afni-glen and @mrneont report that the latest release provides acceptable if not fluid performance on their [MacBooks](https://techable.com/apple/specs/macbook-air-13-inch-core-i5-11-ghz-2020/) that use a [1.1GHz processor with integrated Intel CPU](https://www.notebookcheck.net/Intel-Core-i5-1030NG7-Processor-Benchmarks-and-Specs.458928.0.html) constrained by a 10w TDP and a design with minimal cooling driving a high resolution (2560×1600) monitor.  The pull-request provides further details.

[...]

PR #488, opened 2023-01-22
Slow performance on Intel and Chrome

neurolabusc:
### List of fixed issues

Mitigations for sluggish performance observed on [Intel-based Macs](https://github.com/niivue/niivue/issues/482).

- `meshXRay` is made optional, continuous, and off by default. The [clipplanes](http://192.168.0.220:8080/features/clipplanes.html) demo is updated to illustrate this feature. Rationale: previously there was always a small (2%) `meshXRay`, since this requires the computationally expensive redrawing and blending with [gl.depthFunc(gl.ALWAYS)](https://stackoverflow.com/questions/46216056/side-effects-of-setting-gl-depthfuncgl-always).
- By default, the WebGL context will be created with `antialias: false` if the [hardwareConcurrency](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/hardwareConcurrency) is below a threshold (see below). Rationale: Intel GPUs currently exhibit poor performance with the WebGL default `antialias: true`. This issue is exacerbated on [MacOS computers with `retina displays`](https://stackoverflow.com/questions/30261999/webgl-performance-issues-with-retina-macbook-pro) that have a high pixel count relative to GPU capabilities. As a side note, while HighDPI Windows computers with Intel GPUs seem to work better than Macs (perhaps driver related), Chrome shows a catastrophic performance interaction when you combine `antialias: true` with the developer tools panel being open. This is specific to Windows and not observed on MacOS.
 - Updated [freesurfer](https://niivue.github.io/niivue/features/mesh.freesurfer.html) demo includes `HighDPI` and `MeshXRay` checkboxes to allow developers to see the impact of these choices. By default, it seems we can keep `HighDPI` for Intel machines as long as we switch aliasing off. Rationale: Handling [high DPI](https://www.khronos.org/webgl/wiki/HandlingHighDPI) is computationally expensive, so developers should consider if their application benefits from it.
 - It has been noted that Chrome exhibits [poorer](https://bugs.chromium.org/p/chromium/issues/detail?id=1084591) performance than Firefox in [some](https://monorail-prod.appspot.com/p/chromium/issues/detail?id=1245448) situations. These seem to be related to processing concurrent WebGL frames. This explains why prior versions of NiiVue showed poorer performance on Intel based macs when running Chrome relative to Firefox. Specifically, when mouse move messages are sent faster than the GPU's ability to render a frame. To combat this, a new scene is only rendered once previous frames have completed the asynchronous [gl.finish](https://stackoverflow.com/questions/51710067/webgl-async-operations). The down side is that there may be a tiny bit more frame latency for high performance graphics cards (though still fluid).

The main challenge with implementing this PR is deciding when to implement `antialias: false`. Specifically, it would be great to not penalize users who have computers where the GPU capabilities match the resolution of their screen (with Intel Retina MacBooks being very imbalanced designs). Unfortunately, many web browsers explicitly hide GPU details for the sake of privacy (e.g. I do not know how to discriminate whether Safari is running on an Apple or Intel GPU). This is specific to web pages (Electron can determine these details). This PR implements @hanayik's kludge of using a [hardwareConcurrency](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/hardwareConcurrency) threshold:

 - @hanayik Intel mac reports 4
 - @hanayik Android phone reports 8
- @hanayik iPhone 12 reports 4
- @neurolabusc Intel MacBook 2.3 GHz I5-8259U (MacOS 10.14.6) reports 8
- @neurolabusc Apple M1 MacBookAir reports 8

As can be seen, Apple includes Intel hyper-threading, so we can not clearly discriminate late Intel MacBooks from ARM MacBooks. In theory, an [Intel Macbook](https://en.wikipedia.org/wiki/MacBook_Pro_(Intel-based)) with [9880H CPU](https://ark.intel.com/content/www/us/en/ark/products/192987/intel-core-i9-9880h-processor-16m-cache-up-to-4-80-ghz.html) will report `16` threads and thanks to hybrid graphics could use the Intel GPU to render 3072 × 1920 pixels. 

At the moment, I set `antialias: true` when there are more than [6](https://github.com/niivue/niivue/blob/a55fc2c8d6fc8ec530f7251655bebc1066634f16/src/niivue.js#L547) threads. My hope is that the other mitigations allow this to provide reasonable performance on late Intel MacBooks. But once this PR is accepted, it can be evaluated by @mrneont and @afni-dglen.

### Action Points

 - @hanayik can you accept this PR, as it improves on prior behavior and allows for zero-footprint testing on different devices. Since the thresholding will influence the regression testing bitmaps, I suggest you force this PR, and we can generate new bitmaps once we have agreed on the tradeoff between speed and quality.
 - @mrneont and @afni-dglen once this PR is accepted, can you visit the [live demo page](https://niivue.github.io/niivue/) and see if you prefer the `faster` or `better` option on your Intel-based MacBooks. Please also test both Firefox and Chrome/Safari. My sense is that Firefox WebGL is better for these computers, a situation that is unlikely to be addressed now that Apple is focusing on a different architecture. The `antialiasing` must be chosen when the context is created, so we have separate web pages. However, each has checkboxes that allow you to evaluate how `HighDPI` and `MeshXRay` impacts performance. Finally, with the `Auto-select` web page if you look at the Developer Tools console it will report the number of threads for your device and whether it decided to antialias your context (`NiiVue:  DEBUG AntiAlias  true  Threads  8`).

nv


mrneont:
@neurolabusc : Stunning set of things here. Thanks, and will indeed test when in.

mrneont:
My notes on testing the "Anti-Aliasing: faster (off) or better (on)" triplet of demos here (https://niivue.github.io/niivue/) using my work-supplied Mac laptop with these specs:
* MacBook Air (Retina, 13-inch, 2020)
  macOS Catalina (10.15.7)
  Processor: 1.1 GHz Quad-Core Intel Core i5
  Memory: 16 GB 3733 MHz LPDDR4X
  Graphics: Intel Iris Plus Graphics 1536 MB

... and with these browsers:
* Firefox:  108.0.1 (64-bit)
* Chrome: Version 109.0.5414.87 (Official Build) (x86_64)
* Safari: Version 15.6.1

The testing mainly consists of whirling the brain around at various speeds, using the trackpad, while viewing with the MatCap-Shiny surface on.  I was mainly looking for jumpiness, inconsistency of brain motion or high/changing sensitivity to trackpad motion.
------------------------
**Firefox**
Each of Auto-select, OFF (faster) and ON (better) gave good performance overall at reasonable rotation speed---no obvious jumps.  
I think the OFF (faster) was sliiightly smoother and controlled (and I think the Auto-select matched this).
If I really increase the speed, the "ON (better)" seemed to have a bit more jump.  But that is really only at "plaid speed", as it were, so likely more of a niche case.

**Chrome**
Using ON (better) led to jumpiness even at reasonably low speed.
OFF (faster) and Auto-Select were better and smoother and certainly usable.  I would say that it was not quite as consistent as on Firefox, but these were fine/good performance.

**Safari**
I would say all three methods gave good performance.  At very high speeds ON (better) might be sliiightly poorer than the Firefox performance, but overall it still seemed good.

So, looking quite good across this wide range!  Just "On (better)" on Chrome was less-than-enjoyable.  Thanks!

afni-dglen:
My impressions are a little different from Paul's, but the details are the same. In general, the behavior is still too fast or erratic on my 2020 Intel Mac, but I would say mostly too fast using Chrome. With what I think is a normal dragging speed, I can see about 6 full rotations although it's difficult to count. (This is a little like watching figure skating or gymnastics on TV.) This kind of over sensitivity is also apparent with zooming in and out. The step size for zooms seems too big. So, a few comparisons:

Anti-aliasing - ON or OFF doesn't have a big effect here on the sensitivity, but there is a difference with HighDPI on.

Trackpad vs Logitech mouse. The trackpad is less sensitive, but it's also more difficult to control.  If you're testing with only the trackpad, this might not be an issue.

Matcap vs flat shader - the Matcap is useable with very careful dragging, but very fast. The "flat" shader is too fast most of the time - even on the ARM M1.

HighDPI - better behavior with HighDPI on.

Chrome vs Safari - Safari is a fair bit better for this problem with less over-sensitivity.

ARM vs Intel - The M1 ARM system was useable in every configuration. The flat shader with the mouse was maybe borderline being overly sensitive.

The direction of rotation is very sensitive to the direction of the dragging; if the dragging is not completely horizontal and goes slightly downward, then the direction seems erratic

These tests were with Matcap on using Chrome.
```
              AAON,HD        AAOFF,HD      AAON,HDoff     AAOFF,HDoff
MBAI         Ut,Lm          Ut,Um         Ut,Lm          Ut,Lm
MBAA         Ut,Um          Ut,Um         Ut,Um          Ut,Um
```
MBAI - MacBook Air Intel (2020 version)
MBAA - MacBook Air ARM M1 (2021 version)
AAON - the Anti-Aliasing On "better" solution
AAOFF - the Anti-Aliasing Off "faster" solution
HD - HighDPI on
HDoff - HighDPI off
MatCap - MatCap shading with shiny shader

Issue #510, opened 2023-02-08
是否可以支持vtk文件的渲染


wujie1314:
No examples

hanayik:
@wujie1314 , thanks! Indeed, we seem to have broken our mesh demos due to our recent updates. We will have this fixed soon.

wujie1314:
> @wujie1314 , thanks! Indeed, we seem to have broken our mesh demos due to our recent updates. We will have this fixed soon.

Is the rendering function of vtk files being added?

neurolabusc:
@wujie1314 NiiVue can read [legacy VTK  POLYDATA files described here](http://www.princeton.edu/~efeibush/viscourse/vtk.pdf), we do not support the [VTK XML format](https://kitware.github.io/vtk-examples/site/VTKFileFormats/). We have also tried to support some of the [undocumented features of legacy VTK format](https://discourse.vtk.org/t/upcoming-changes-to-vtkcellarray/2066). Our primary aim was to support the variations popular in our domain: tractography VTK streamlines (saved with either the .vtk or .fib extension) and improve Slicer3D interoperability. The VTK format was really designed to showcase VTK features, and is less well defined than many of the formats we support. 

If you want to test NiiVue's VTK support, open a mesh with [Slicer3D](https://download.slicer.org/) and choose to save it as `Poly Data (.vtk)`. You can then drag and drop this onto NiiVue. Note that NiiVue explicitly **does not** support VTK voxel-based data due to its inability to encode spatial transforms.
![slicer](https://user-images.githubusercontent.com/8930807/219801694-0d266736-35aa-4678-ae61-56e3d96003c1.png)
![niivue](https://user-images.githubusercontent.com/8930807/219801709-a992ddd4-29f8-4906-91ee-446d01a059e7.png)

PR #512, opened 2023-02-08
Issue 510

hanayik:

- #510


hanayik:
Looks like these snapshots are failing because meshes are not being rendered. They will also fail due to UI widget differences between OSes captured in the full page screenshots. I will need to just save the image of the canvas (like our other tests).

neurolabusc:
Yes, your patch looks more comprehensive, but the minimal fix is [to change one line](https://github.com/niivue/niivue/blob/ded48d564265824210a00e28f1901f786941f4ca/src/nvmesh.js#L3593) from:
```
//urlParts = layer.url.split("/");
```
to define this this variable, e.g.
```
let urlParts = null;
```

hanayik:
@neurolabusc , thanks for the minimal code, but I do not think that is impacting the tests, because they work locally. 

There's a few issues here:
1. I was attempting to create screenshots of our demos pages for meshes, so that we avoid publishing broken demos in the future. 
2. however, the demos pages will look different based on the OS (standard sliders, etc. in html look different on Mac vs. Win). I will need to just capture the canvas image, rather than the full page screenshot to avoid the UI widgets. 
3. the mesh demos don't seem to render in headless mode. @hanayik can test this more on a windows machine that will do headless tests.

Issue #524, opened 2023-02-22
render views not interactive via mouse or touch input


hanayik:

neurolabusc:
@cdrake this was an unintended consequence of the patch for [issue 518](https://github.com/niivue/niivue/issues/518). Rotating a 3D volume does not change the cross-hair position in image or world space, so we should not generate a location change event. Rather, we generate location change events when a user double-clicks on a volume. Happy to have a Zoom meeting if my patch reintroduces your issue. It is just unclear to me what your issue was and how to replicate it.

cdrake:
Location change event is for the cursor generated by both 2D and 3D views. Bug introduced because I made a mistake in order of assignment.  Will provide fix (with tests) tonight.

PR #526, opened 2023-02-22
Thresholding

neurolabusc:

- [issue 524](https://github.com/niivue/niivue/issues/524)


hanayik:
We should also make a new test for this so that we make sure this doesn't happen in the future. 
the test should:
- load a volume in 3d mode
- click and drag the mouse a few pixels
- save a screen shot of the canvas

should be similar to the right click and drag demo.

PR #545, opened 2023-03-15
Desktop gui features

hanayik:


src/cmaps/index.js
```diff
@@ -1,5 +1,5 @@
-export { default as _itksnap } from "./_itksnap.json";
-export { default as _slicer3d } from "./_slicer3d.json";
+export { default as itksnap } from "./_itksnap.json";
+export { default as slicer3d } from "./_slicer3d.json";
```
neurolabusc, line 2: Please note that the underscore prefix for _itksnap and _slicer3d denote that these are indexed colormaps for drawing pens, to distinguish them from interpolated/continuous colormaps used for scalar voxel data. If you change this, some demos will list poorly suited color schemes, e.g. 
  https://niivue.github.io/niivue/features/colormaps.html
I am fine if you want to use a different method of discriminating these (e.g. a ‘cmaps’ folder for colormaps and ‘maps’ for drawing colors), but you will need to update the demos and my standalone software.

src/cmaps/index.js
neurolabusc, line 1: If you change '_' to '$' in /cmaps/index.js make sure to update [colortables.js](https://github.com/niivue/niivue/blob/2e8343274f92ccee24ac6df6eb8634437de93de7/src/colortables.js#L11) to detect this.

PR #550, opened 2023-03-27
add an ROI-based colormap from AFNI

mrneont:

-
-
-


neurolabusc:
You should change:
```  
"min": 0,
  "max": 255,
```
to read
```  
"min": 0,
 "max": 0,
```
This refers to the voxel intensity for the first and last index in your colormap. Since MR scans have relative intensities, you should not specify this. NiiVue will default to the [robust range](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/Fslutils) for voxel intensities unless the NIfTI header or color table specify otherwise (and the user can interactively adjust this by dragging over a 2D slice).

hanayik:
@mrneont, is this PR still a work in progress?

mrneont:
@hanayik : I (think I) am happy with those cmaps if you are!  Those should be the main ones for the APQC HTML, at present (you already have a Plasma cmap in the distribution).

neurolabusc:
Looks good to me.

Issue #501, opened 2023-02-01
Only 2 gz files can be displayed when the latest version 32 is loaded


wujie1314:
With BUG
Only 2 gz files can be displayed when the latest version 32 is loaded

neurolabusc:
Unable to replicate. This script shows 4 voxel based images overlaid with NiiVue:

```
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>NiiVue</title>
  </head>
  <body>
    <canvas id="gl" width="1024" height="640"></canvas>
  </body>
<script src="https://unpkg.com/@niivue/niivue@0.32.0/dist/niivue.umd.js"></script>
<script>
 var root = "https://niivue.github.io/niivue-demo-images/"
 var volumeList = [
  {url: root+"chris_t1.nii.gz", colorMap:"gray"},
  {url: root+"mni152.nii.gz", colorMap:"blue", opacity:0.2},
  {url: root+"spmMotor.nii.gz", colorMap:"red", opacity:0.5, cal_min: 1, cal_max: 5},
  {url: root+"spmMotor.nii.gz", colorMap:"green", opacity:0.5, cal_min: 4, cal_max: 5},
 ]
 var nv = new niivue.Niivue({isResizeCanvas: false})
 nv.attachTo("gl") 
 nv.loadVolumes(volumeList)
 nv.opts.isColorbar = true
</script>
```

![overlay](https://user-images.githubusercontent.com/8930807/216015865-90c94289-4493-408f-9d29-d8621e6cb81a.jpg)

Perhaps you can provide a script that illustrates your issue.

neurolabusc:
In investigating this issue, I did notice that adding overlays can restrict the bounding box of previous overlays. While it is typical that all overlays share a common bounding box (and we expect that all voxel overlays are constrained by the bounding box of the background image), I have relaxed the constraint between overlays.

The script below illustrates the previous behavior, where the (misregistered) green overlay clips the signal of the red overlay.

```
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>NiiVue</title>
  </head>
  <body>
    <canvas id="gl" width="1024" height="640"></canvas>
  </body>
<script src="https://niivue.github.io/niivue/features/niivue.umd.js"></script>
<script>
 var root = "https://niivue.github.io/niivue-demo-images/"
 var volumeList = [
  {url: root+"spm152.nii.gz", colorMap:"gray"},
[...]

PR #581, opened 2023-05-01
FIX: Check if rendererInfo is non-null before looking up properties

### LINKED ISSUE #582: Catch is resistFingerprinting is set
The Mozilla FireFox browser will block NiiVue if `privacy.resistFingerprinting` is set to true, as these lines that attempt to detect details regarding the graphics capabilities are considered a privacy issue:

```
let vendor = this.gl.getParameter(rendererInfo.UNMASKED_VENDOR_WEBGL);
let renderer = this.gl.getParameter(rendererInfo.UNMASKED_RENDERER_WEBGL);
```

To replicate, enable this feature: 
1. Type `about:config` in the address bar and press Enter. A warning page may appear. Click Accept the Risk and Continue to go to the `about:config` page.
2. Search for privacy.resistFingerprinting and set it to `true`.

effigies:
The `WEBGL_debug_renderer_info` extension produces a deprecation warning in Firefox and causes an error if `privacy.resistFingerprinting` is enabled in Firefox config. This change simply checks whether `rendererInfo` evaluates `true` before trying to look up properties in it.

Resources used to propose this:

* https://bugzilla.mozilla.org/show_bug.cgi?id=1722782
* https://stackoverflow.com/questions/69055946/what-does-the-message-webgl-debug-renderer-info-is-deprecated-in-firefox-and-wi
* https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/Constants#getting_gl_parameter_information

Closes #582.


neurolabusc:
@effigies We recognize that Mozilla/FireFox intentionally masks this information, however these details can help debug data from other browsers. This explains why we report ***both*** the `renderer` and the `firefox` renderer:

```
  let rendererInfo = this.gl.getExtension("WEBGL_debug_renderer_info");
  let vendor = this.gl.getParameter(rendererInfo.UNMASKED_VENDOR_WEBGL);
  let renderer = this.gl.getParameter(rendererInfo.UNMASKED_RENDERER_WEBGL);
  log.info("renderer vendor: ", vendor);
  log.info("renderer: ", renderer);
  //firefox masks vendor and renderer for privacy
  let glInfo = this.gl.getParameter(this.gl.RENDERER);
  log.info("firefox renderer: ", glInfo); //Useful with firefox "Intel(R) HD Graphics" useless in Chrome and Safari "WebKit WebGL"
```

This content is only there to provide debug for users who have issues, we could always simply remove all of these lines. At the moment, we use `navigator.hardwareConcurrency` to try to decide whether to enable AntiAliasing (which is nicer but slower). The number of CPU threads is a poor surrogate for GPU performance, but as you note some browsers consider even basic GPU details as private.

A great feature of WebGL2 is that it uses the best available API (OpenGL ES, Vulkan, Metal, or DirectX). However, this means that we sometimes need to debug bugs associated with specific implementations, for example see [here](https://github.com/hanayik/niivue/issues/83), [here](https://github.com/niivue/niivue/issues/344) and [here](https://github.com/niivue/niivue/issues/501).

effigies:
Ah, I should have read further down. My javascript is poor enough that I have a bit of tunnel vision. Could this be wrapped in something like a `try/catch` to avoid crashing when `privacy.resistFingerprinting` is enabled?

neurolabusc:
@effigies your updated code looks perfect. @hanayik can you accept this PR to [close issue 582](https://github.com/niivue/niivue/issues/582).

PR #585, opened 2023-05-09
niivue.js: when modulating image (alpha transparency), user sets powe…

mrneont:
…r (def = 2, a l'AFNI)

- if applying a modulation image, allow more flexible alpha scaling than linear
- user controls with a kwarg in setModulationImage()
- set default scaling to quadratic, matching AFNI


neurolabusc:
@mrneont in general this looks fine.
 1. I wonder if we should have **both** the boolean  **modulateAlpha** as well as the float **modulatePower**. An alternative might be to have a single float **modulateAlphaPower** that is set to **NaN** if an image is not using modulation. I tend to prefer a single parameter versus two.
 2. When introducing new features, we should also ensure that the live demos illustrate their usage. I would suggest updating the [modulateScalar](https://niivue.github.io/niivue/features/modulateScalar.html) demo to show the new power function. Our documentation links to these live demos, so it gives developers a concrete example of usage.

mrneont:
@neurolabusc : that is fine---whichever design choice you think is easier.  

I might suggest adding in the docs to note to people that the quadratic scaling might be the best place to start for a more "highlighting" aspect, esp. at the single subject level.

mrneont:
This PR could be removed.  It is superceded by: https://github.com/niivue/niivue/pull/588

Issue #608, opened 2023-05-29
Diverging colors and alpha thresholding


neurolabusc:
Last weeks nilearn meeting discussed several desired features for displaying statistical maps on meshes.

1.  Several [matplotlib](https://matplotlib.org/stable/tutorials/colors/colormaps.html) colormaps are diverging. It might be nice that as a statistical value became less extreme, it became a more neutral color. I have added the new `green2orange` and `green2cyan` color maps that both start at a neutral dark forest green and add red or blue to signify more extreme values.
2. Our voxel-based statistical maps allow the `alphaThreshold` which determines if the color range starts from zero or from the least extreme value. Having this option is useful, so it should exist for meshes.

I have created a pull request that implements both features, and these are showcased in a new `mesh statistics` live demo:

![diverge](https://github.com/niivue/niivue/assets/8930807/1e99a3a8-ce3a-441c-8b00-344914c0f074)

PR #609, opened 2023-05-29
Diverging colors, mesh alphaThreshold

neurolabusc:

- [issue 608](https://github.com/niivue/niivue/issues/608)


hanayik:
@neurolabusc , I suspect the tests will pass now. I ran `npm run test` locally to generate the new snapshots for additional colormaps. Now the GitHub actions job will not fail due to not having those particular snapshots.

hanayik:
@neurolabusc , the windows tests seem to be failing (not related to this PR). They don’t seem happy with the AFNI file loader test for some reason:

https://github.com/niivue/niivue/actions/runs/5136979649/jobs/9244435051?pr=609#step:7:165

I suspect this was always the case but we didn’t know until the file loader test suite was expanded. I suggest we merge this PR in now, and submit a new issue and PR for the AFNI file loader tests fix on windows.

neurolabusc:
@hanayik remember that web pages live in a sandbox, so for dual-file formats like HEAD/BRIK you need to specify ***both*** files. [Here us an example](https://niivue.github.io/niivue/features/afni.html):

```
      url: "../images/example4d+orig.HEAD", //"./images/RAS.nii.gz", "./images/spm152.nii.gz",
      urlImgData: "../images/example4d+orig.BRIK.gz",
```

The [test.fileFormats.js](https://github.com/niivue/niivue/blob/fd1d35ce853b41e35860d67d1de6a847e47614d7/tests/test.fileFormats.js#LL42C9-L42C12) is only specifying the `url`, not the `urlImgData`.

hanayik:
We treat HEAD as a special case when loading from URL and automatically search for BRIK at the same location. 

https://github.com/niivue/niivue/blob/fd1d35ce853b41e35860d67d1de6a847e47614d7/src/nvimage.js#L2786

and that doesn’t explain why it only fails on windows. Perhaps related to the line-by-line parsing of the HEADer? I’m on mobile, but can investigate more later.

PR #634, opened 2023-06-27
Feature/1322 voxel edit undo redo

MelGGit:

- added functionality for undos and redos
- implementation of stack class


MelGGit:
@neurolabusc Thank you a lot for your input. I was on a conference the last week so I could not reply last week. I tried to get an image which uses a 32-bit RGBA was wasn't able to. Do you have a scenario where this would apply?
I noticed when using the scl_slope and scl_inter the brightest value would not be as bright as the brightest value available in the array. This would mostly apply to Int8Arrays. So in most IntArrays the highest value or brightest value would be 255 but if I include the scl_slope, I receive 91,... as highest value. Do you have a solution in mind for that?

neurolabusc:
NiiVue can show 32bit RGBA images (packed as RGBA quads in a Uint8Array). NiiVue [data scaling](https://brainder.org/2012/09/23/the-nifti-file-format/) uses `scl_slope` and `scl_inter` to convert raw image data into calibrated units. Hence the if your raw uint8 data is in the range of 0..255 but your `scl_slope` transforms this to a calibrated 0..91, your effective range for `cal_min` and `cal_max` are 0..91.

Here is the example visible human RGBA image that comes with MRIcroGL:

visible

Issue #640, opened 2023-07-04
Shiny volume rendering


neurolabusc:
[MRIcroWeb](https://github.com/rordenlab/MRIcroWeb) was the forerunner to NiiVue and included gradient-based lighting for the volume rendering. This method requires more video memory and the rendering is a bit slower, so this should be an option rather than the default behavior.

![shiny](https://github.com/niivue/niivue/assets/8930807/c826e42c-8cc3-4c3f-b6ab-6978b5b1f46e)

PR #641, opened 2023-07-04
Shiny volume rendering

neurolabusc:
List of fixed issues:

- https://github.com/niivue/niivue/issues/640 with live demo.
- This should be fully functional, but not yet added to the document format. This mode is optional, as it requires more resources.


hanayik:
@neurolabusc, the shiny light source in this commit doesn't seem to mimic the MRIcroGL or MRIcroWeb behaviour. For example, I can't get any shiny surface on the right lower 2/3rds of this [brain extracted image](https://github.com/niivue/niivue/files/11957382/T1_brain.nii.gz) (screenshot attached) when rotating it in the scene. 

## this commit
image

## MRIcroGL
image

## MRIcroWeb
image

neurolabusc:
@hanayik good catch. I have fixed the view-dependent shading. I have also allowed the shader to use the same matcap as the meshes, and also made it so the clip planes do not glisten with the selected lighting.

matcap_volume

PR #650, opened 2023-07-18
add gray_inv colorbar

mrneont:

-
-
-


hanayik:
@mrneont and @neurolabusc , I think it might be nicer to just have a new Niivue function that reverses a given colormap. That idea was originally suggested by @mrneont , but he and I recently used the custom colormap functionality in Niivue to apply "gray_inv" as a colormap to a volume in https://github.com/afni/afni/pull/380

neurolabusc:
@hanayik I also like the idea of an option to reverse the color maps. Does this reverse just the red, green, blue color components, or also the alpha? In other words, it is traditional in volume rendering for darker intensities to be more transparent than brighter values. My sense (and the approach of the gray_inv submitted) is that the order of R,G,B nodes are mirrored but not the Alpha.

mrneont:
I agree that just reversing RGB channels, but leaving the A and I ones unchanged, makes sense.

PR #682, opened 2023-09-04
Feature save html

cdrake:

-[Save as HTML](https://github.com/niivue/niivue/issues/680) (uses default HTML)
-Save as Template HTML (use template HTML with %%javascript%% for document loading script replacement) 
-Opts and Scene data was not being saved


src/niivue.js
```diff
@@ -3254,11 +3258,153 @@ Niivue.prototype.loadDocument = function (document) {
...
+  delete json.sceneData["onAzimuthElevationChange"];
+
+  let docString = JSON.stringify(json);
+  const buf = fflate.strToU8(docString, { level: 6, mem: 4 });
+  const compressed = fflate.compressSync(buf);
+
+  // https://stackoverflow.com/questions/68849233/convert-a-string-to-base64-in-javascript-btoa-and-atob-are-deprecated
+  const base64 = NVUtilities.uint8tob64(compressed);
+  const javascript = `
+  import * as niivue from "https://niivue.github.io/niivue/features/niivue.es.js";
```
hanayik, line 3294: @cdrake , Ideally it would be nice to avoid depending on CDN imports for the saved HTML. The saved HTML files will depend on an internet connection to display the contents right?

It's possible we could add a new `define` key [here](https://github.com/niivue/niivue/blob/3c3fd2bd836deb028c6737e065e86d8bbf9bbac6/vite.config.js#L7) that would inject the compiled niivue bundle (either the .es or .umd file). This will bloat the final output (double in size) but will allow niivue to inject its own code without a CDN. 

For `fflate` could we just have an alias fflate call in niivue (`Niivue.prototype.<insert_fflate_function>`) that utilises the fflate we are already importing as part of niivue?

hanayik (changes requested):

tests/test.saveImageDocument.js
```diff
@@ -0,0 +1,42 @@
...
+    // SAGITTAL: 2
+    nv.setSliceType(2);
+    const json = nv.json();    
+    let document = niivue.NVDocument.loadFromJSON(json);
+    
+    
+    return document;
+  });
+
+  console.log(document);
```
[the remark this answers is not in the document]
cdrake: removed

src/niivue.js
cdrake: Sounds good to me. I think it would be nice for users because the documents would stay functional even with major changes to NiiVue.

tests/__image_snapshots__/test-load-document-js-load-document-1-snap.png
hanayik, line 1: @cdrake , is this new snapshot for loadDocument accurate? Just checking since it is very different (3D vs 2D multiplanar). If we were not loading scene and opts variables correctly anyways, I guess your new snapshot shows that we do that correctly now?

hanayik (changes requested):

tests/__image_snapshots__/test-load-document-js-load-document-1-snap.png
cdrake: Yes. When the niivue opts and scene were decoupled from the document, the load/save document methods were neither  loading nor saving that data.

hanayik:
@cdrake , I will run the windows tests locally when I get time today (7th sep) and see if I can replicate what is happening in the github actions machines.

hanayik (approved):

PR #686, opened 2023-09-07
add two stage build to inject niivue UMD

hanayik:

- related to #682 and #680 
- added a new vite config file to embed the previously bundled niivue UMD (now a two stage build process)
- the new build process does this:

1.   first pass builds niivue.es.js and niivue.umd.js
2. second pass reads the contents of niivue.umd.js via the new vite config file and injects the base64 encoded, compressed version of niivue.umd.js
3. The build is actually not as large as estimated, so the size cost of doing this is negligible 

- added a new function to niivue: `decodeEmbeddedUMD()`. This function will decode the base64 string and decompress the embedded Niivue UMD bundle so that the string contents can be injected into HTML files that Niivue saves.
- To try it, you can put `console.log(this.decodeEmbeddedUMD());` inside `attachToCanvas` to see the contents of the UMD version of niivue printed in the console.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .DS_Store
 node_modules
 /dist
+dist_intermediate
 devdocs
 /_md_docs
 /docs/niivue*.js
--- a/package.json
+++ b/package.json
@@ -13,8 +13,8 @@
   },
   "scripts": {
     "dev": "vite",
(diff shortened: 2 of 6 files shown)


cdrake (approved):
Looks good to me.  Added a test to verify ending of UMD string.  Feel free to remove.

PR #687, opened 2023-09-07
MHA reader fixes

EdwinBennink:
List of fixed issues:
- MHA reader failed on reading text files with DOS line endings.
- 2D MHA images could not be read because the reader expected a 3rd dimension.
- The MHA orientation matrix was incorrectly scaled.
- For compressed paired images, the reader tried to decompress the main buffer instead of the paired image.

### DIFF
--- a/src/nvimage.js
+++ b/src/nvimage.js
@@ -304,7 +304,7 @@ export function NVImage(
       break;
     case NVIMAGE_TYPE.MHD:
     case NVIMAGE_TYPE.MHA:
-      imgRaw = this.readMHA(dataBuffer); //to do: pairedImgData
+      imgRaw = this.readMHA(dataBuffer, pairedImgData);
       break;
     case NVIMAGE_TYPE.MGH:
     case NVIMAGE_TYPE.MGZ:
--- a/src/nvimage.js
+++ b/src/nvimage.js
@@ -1465,17 +1465,21 @@ NVImage.prototype.readMHA = function (buffer, pairedImgData) {
     throw new Error("File too small to be VTK: bytes = " + buffer.byteLength);
   var bytes = new Uint8Array(buffer);
   let pos = 0;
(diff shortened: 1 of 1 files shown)


neurolabusc:
@EdwinBennink can you provide sample [MHA format](https://itk.org/Wiki/ITK/MetaIO/Documentation) images that exhibit DOS `\r\n` rather than UNIX `\n` end of line encoding? I have never seen one in the wild.

EdwinBennink:
Hi @neurolabusc, [this MHD/ZRAW pair](https://github.com/niivue/niivue/files/12577426/hand.zip) has DOS line endings, is 2D and uses compression in the paired ZRAW file. It is also the first time I encountered an MHA image with `\r\n` endings, but I could not find any restrictions on the line endings in the [specifications](https://itk.org/Wiki/MetaIO/Documentation) either, so I would not consider it to be 'wrong'.

hanayik:
@EdwinBennink and @neurolabusc  , have you successfully tested that the example MHD/ZRAW file can be read and rendered by NiiVue (I have tested in ITK SNAP and it works)? I have not been able to successfully render the image in my brief trials using the `Edwinbennink:mha-reader` branch. It would also be nice to add a test for this use case so that we can ensure we don't break things during future development. I'm happy to help add the test if you can help make sure the offending image pair is renderable in NiiVue.

EdwinBennink:
Hi @hanayik, I pushed the changes to the readMHA function, but did not push the necessary change to the NVImage class in my first commit (to actually call `readMHA` with the `pairedImgData` argument). I did that a moment ago; does it now work for you?

hanayik:
@EdwinBennink , yes, looks good with the demo data you provided :). thanks! I will close this PR and merge in the one that I opened here that incorporates everything plus a bump in the version number #697

PR #688, opened 2023-09-07
Fix DICOM scaling

EdwinBennink:
List of fixed issues:
- The DICOM slice thickness was used as pixel depth, but the pixel depth should be calculated using the offsets between subsequent slices. The slice thickness is frequently larger than the slice spacing, and in case of tilted slices (used in helical CT's, for example), the pixel depth is larger than the slice spacing because of the skew.

### DIFF
--- a/src/nvimage.js
+++ b/src/nvimage.js
@@ -953,10 +953,18 @@ NVImage.prototype.readDICOM = function (buf) {
   let rc = this.series.images[0].getPixelSpacing(); //TODO: order?
   hdr.pixDims[1] = rc[0];
   hdr.pixDims[2] = rc[1];
-  hdr.pixDims[3] = Math.max(
-    this.series.images[0].getSliceGap(),
-    this.series.images[0].getSliceThickness()
-  );
+  if (this.series.images.length > 1) {
+    // Multiple slices. The depth of a pixel is the physical distance between offsets. This is not the same as slice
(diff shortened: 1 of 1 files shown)


neurolabusc:
@EdwinBennink I have reviewed your suggestion and it looks sound. The rationale is [described here](https://neurostars.org/t/calculating-voxel-resolution-from-dicom-headers/7091/2). 

@Hanayik we should include a regression test that includes a DICOM series with a spatial gap between slices. We will do everything we can to support DICOMs. We do have a [dcm2niix validation dataset that includes gantry tilt in the DICOMs](https://github.com/neurolabusc/dcm_qa_ct). I am not sure if Daikon handles these correctly, but dcm2niix does and we [provide a live demo to illustrate NiiVue's gantry tilt support](https://niivue.github.io/niivue/features/shear.html) if it is correctly reported in the NIfTI SForm.

However, our niche is to support neuroimaging formats (voxels, meshes, connectomes, statistical maps, tractography). I think potential developers should consider dedicated DICOM viewers like [OHIF viewer](https://viewer.ohif.org/) if they want a focused DICOM tool.

neurolabusc:
@hanayik I approve this PR. Please merge this when convenient. I think the best description is [this figure](https://github.com/rordenlab/dcm2niix/issues/253#issuecomment-449465111).

Issue #695, opened 2023-09-12
Dragging between tiles


neurolabusc:
NiiVue can show multiple image tiles simultaneously, e.g. coronal, sagittal, axial and rendering views can be shown in the same WebGL context. NiiVUe also has dragging gestures to pan/zoom, set contrast, measure dimensions, etc. Unfortunately, one can get unexpected results if one starts a drag in one tile and drags the mouse to another. The drag gestures should be constrained to the tile where the drag originated.

PR #696, opened 2023-09-12
Constrain drags to initial tile 

neurolabusc:

- #695 this makes sense for all gestures, with the possible exception of the panning gesture.


neurolabusc:
Also resolves https://github.com/niivue/niivue/issues/700

Issue #706, opened 2023-09-23
Focus should determine if a canvas will respond to a sync message


neurolabusc:
The current [sync live demo](https://niivue.github.io/niivue/features/sync.mesh.html) has one master image and one slave. This is a one-directional yoking, as changes in the slave are not reflected to the master. I would suggest that syncing messages should only be sent with the canvas that has focus. This would allow multiple canvas to each control each other without problems with infinite loops. 
![bidirectionalSync](https://github.com/niivue/niivue/assets/8930807/bc43f4bf-4c1a-414e-9453-ccb9cf2b1c4b)

PR #708, opened 2023-09-28
Minor fixes

neurolabusc:

- #705
- #706 : note that this PR allows bidirectional connections between two NV instances. But to fully resolve this issue I would allow syncWith() and sync() to allow otherNV to either be one NiiVue instance or an array of NiiVue instances that would allow more than a pair of yoked images.


hanayik:
added `broadcastTo(object | array)` and bumped version

PR #738, opened 2023-11-16
Feature graceful mz3 load

cdrake:

-Handle mz3 file loading with double value overlays
-Add tests for loading mz3 files
-


hanayik:
Nifti-reader-js is broken due to a packaging error. I will fix that and release a new version, and then bump the version in niivue which will fix the tests here that are failing.

PR #741, opened 2023-11-16
Use Matrix action for testing

jens-ox:
I wasn't really sure why this is the case, but all the testing actions (`test-x-y.yml`) declare a matrix but don't use it - all of them only specify one value for `node-version`, defying the purpose of the `matrix` keyword.

This PR reduces all those actions into one single one, moving the node version and the OS to a `matrix`. The testing failure is the same as on the current main.

Additionally, the following things have been done:
- Removed `rm package-lock.json` - the lockfile exists to ensure reproducibility of tests in CI.
- Added `puppeteer-core` to `devDependencies` to avoid having to install it as a separate step in CI.
- Updated all consumed workflows to their latest major version.

Edit: Sorry for the big lockfile diff, it seems like the lockfile was out of date.


hanayik:
Nifti-reader-js is broken due to a packaging error. I will fix that and release a new version, and then bump the version in niivue which will fix the tests here that are failing.

jens-ox:
@hanayik I rebased and [the run was successful](https://github.com/jens-ox/niivue/actions/runs/6901929524) ☺️

PR #742, opened 2023-11-17
Jens ox main

hanayik:


jens-ox:
Hi @hanayik! As this PR is identical to #741, you can probably close the other one. Alternatively, you can approve the workflow run on the other PR to allow merging. The advantage would be the better papertrail in the other PR ☺️

Issue #739, opened 2023-11-16
nvimage: unclear code


jens-ox:
In [`nvimage.js`](https://github.com/niivue/niivue/blob/main/src/nvimage.js#L1908), some switch statements are a bit hard to understand due to declarations within cases (which are [not lexically scoped](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#lexical_scoping) and missing break statements.

I don't have enough understanding of the code to raise a PR fixing this, so it would be great if someone more knowledgeable than me could have a look ☺️

neurolabusc:
This function readNRRD() provides a direct mapping from the [NRRD](https://teem.sourceforge.net/nrrd/format.html#spacings) fields to the [NIfTI](https://nifti.nimh.nih.gov/pub/dist/src/niftilib/nifti1.h) fields as NiiVue uses NifTI internally for its voxel-based format. Therefore, the line you link to directly maps the NRRD array `spacings[]` to the NIfTI array `pixdim[]`.  Perhaps because I am so familiar with these formats, this seems like the most concise method to convert this format. Please provide guidance to how this can be simplified.

jens-ox:
Hi @neurolabusc! Thanks for the quick reply.

I would change the following:
- add a `break` statement to `case "spacings"`
- scope all declarations in `case "spacings"` and `case "sizes"`
- if it's intentional that `case "spacings"` also executes `case "sizes"`, change the code accordingly

The resulting code would look something like this:

![CleanShot 2023-11-16 at 14 40 26@2x](https://github.com/niivue/niivue/assets/5368029/f07521c5-3cd9-4f58-9ca4-7d57c657bd42)

Assuming that the current behaviour is correct and that `case "sizes"` should indeed run for `spacings`, the code would look like this:

![CleanShot 2023-11-16 at 14 43 04@2x](https://github.com/niivue/niivue/assets/5368029/7c3db32a-142a-4b99-8f58-a25d1567c10f)

neurolabusc:
@jens-ox mea culpa, break added.

jens-ox:
@neurolabusc this will likely still cause issues, as `dims` will not be properly defined.

If it's ok for you I'll raise a PR to introduce some basic linting that chatches stuff like this automatically.

PR #743, opened 2023-11-17
Linting setup

jens-ox:
Hi all!

This PR updates the linting setup to use a stricter ESLint configuration. Specifically, it integrates Prettier into ESLint (which was already installed but was invoked by the build script) and uses the [JavaScript Standard Style](https://standardjs.com/). I also added linting to the CI action.

This should catch a couple of common gotchas, e.g. #739.

The second commit of this PR includes all changes caused by running `eslint --fix .`, reducing the amount of "errors" from ~16.5k to ~1k. The remaining "errors" require manual work, which I would do if the changes to the linting setup introduced by this PR are ok for everyone.


jens-ox:
@hanayik @neurolabusc does this linting setup make sense for you? If the config is ok, I would start addressing the linting issues that need manual fixing.

neurolabusc:
Looks good to me. @hanayik has been maintainer of these features so may have strong opinions.

jens-ox:
@neurolabusc @hanayik I tried to separate the different types of issues into different commits. I marked the non-trivial ones with `[nontrivial]` - those are probably the only commits worth looking at.

13 problems remain that require more knowledge of the code than I have, all are in `nvmesh-loaders.js`. It would be great if someone could have a look at them, this PR is ready otherwise.

CleanShot 2023-11-19 at 15 44 03@2x

jens-ox:
@neurolabusc could you approve the workflow run?

hanayik (commented):
@jens-ox , by switching to `hasOwn` aren't we dropping [support](https://caniuse.com/?search=hasOwn) for many browser versions released prior to 2021/2022 (based on the browser)? For example, we support Safari `15`+ because it added WebGL2 support without being an experimental flag, but `hasOwn` was introduced in Safari `15.4`, so that would now be the lowest supported version right?

hanayik (approved):

hanayik (changes requested):
@jens-ox, there's a problem with loading some image formats that I need to track down. For example, the image `<project_root>/tests/images/FLAIR.nrrd` won't load via drag and drop, along with the `enh.dcm` image in the same directory. They load ok from URL, just not drag and drop

jens-ox:
@hanayik 
> by switching to hasOwn aren't we dropping [support](https://caniuse.com/?search=hasOwn) for many browser versions released prior to 2021/2022 (based on the browser)? For example, we support Safari 15+ because it added WebGL2 support without being an experimental flag, but hasOwn was introduced in Safari 15.4, so that would now be the lowest supported version right?

I reverted this again in https://github.com/niivue/niivue/pull/743/commits/1ddb5fde931ce065f1a12f35c1cc076175840d22, has `hasOwn` (or `hasOwnProperty`) wasn't really needed. Those comparisons can be further simplified once a port to TS starts.

> there's a problem with loading some image formats that I need to track down. For example, the image <project_root>/tests/images/FLAIR.nrrd won't load via drag and drop, along with the enh.dcm image in the same directory. They load ok from URL, just not drag and drop

How can I test this?

hanayik:
@jens-ox , yep, unfortunately there are no tests for drag and drop file loading (could just as easily be a file input element in testing). I’ll keep investigating this during the morning.

jens-ox:
@hanayik I also investigated for a bit this morning but wasn't able to resolve it. Should this be fixed in this PR or in a follow-up?

hanayik:
Since the existing tests pass, and this PR is already large enough, I’ll merge it now, and then address the file loading issue via drag and drop

PR #745, opened 2023-11-21
colortables.js - convert to ES2015 class

jens-ox:
> This is a follow-up PR to #743, so this should remain a draft until that PR is merged.

In order to facilitate a TS migration in the future, it makes sense to modernise the codebase first. One obvious step is to move the old-school function-based classes to "proper" ES2015-style classes ([see MDN entry](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes)).

This PR moves `colortables.js` to ES2015 classes.

### DIFF
--- a/src/colortables.js
+++ b/src/colortables.js
@@ -1,222 +1,226 @@
 import * as cmaps from './cmaps'
 
-export const colortables = function () {
-  this.version = 0.1
-  this.gamma = 1.0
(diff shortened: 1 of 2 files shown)


jens-ox:
Closing in favor of one PR to convert the codebase to ES2015 classes.

Issue #755, opened 2023-11-25
avoid omitting curly brackets in block statements


hanayik:
NiiVue contains hundreds of block statements that omit curly brackets. The code is more readable if we include the brackets (And force this using a new [eslint rule](https://archive.eslint.org/docs/2.0.0/rules/curly))

PR #756, opened 2023-11-25
added curly config

hanayik:

- #755 

added new curly config to eslint and run `npx eslint --fix .` to update all block statements with curly brackets.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,10 +4,11 @@ module.exports = {
     jest: true
   },
   root: true,
-  ignorePatterns: ['dist/**/*'],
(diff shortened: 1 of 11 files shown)


hanayik:
@neurolabusc, @cdrake, @jens-ox: do you have any strong opinions about this new eslint rule? I personally think curly brackets should always be used. I figured this cosmetic change was due since many other code style enhancements have been added recently.

jens-ox (approved):
This makes a lot of sense! 👍🏼

neurolabusc:
@hanayik this is not my personal style: I use curly braces to denote scope, so automatically universally adding them makes it harder for me to infer intention. However, this is a small concern and I am happy to live with the consensus that others find useful.

PR #767, opened 2023-11-29
Fix mesh visible

### LINKED ISSUE #762: What does mesh.visible do?
I am trying to develop an application where the visibility of meshes can be toggled. I see that meshes have a field called `visible` which can be set by calling `nv1.setMeshLayerProperty`.

```patch
diff --git a/demos/features/mesh.layers.html b/demos/features/mesh.layers.html
index e18081b..66c39c3 100644
--- a/demos/features/mesh.layers.html
+++ b/demos/features/mesh.layers.html
@@ -33,6 +33,13 @@
         class="slider"
         id="meshSlider3"
       />
+      <input
+        type="checkbox"
+        id="meshCheckbox1"
+        checked
+      />
+      <label for="meshCheckbox1">Visible</label>
+
     </header>
     <main id="container">
       <canvas id="gl1"></canvas>
@@ -52,6 +59,11 @@
   slider3.oninput = function () {
     nv1.setMeshLayerProperty(nv1.meshes[0].id, 0, "opacity", this.value * 0.1);
   };
+  var checkbox1 = document.getElementById("meshCheckbox1");
+  checkbox1.onclick = function () {
+    nv1.setMeshProperty(nv1.meshes[0].id, 'visible', checkbox1.checked);
+    console.log(`visible=${nv1.meshes[0].visible}`);
+  }
   function handleLocationChange(data) {
     document.getElementById("location").innerHTML =
       "&nbsp;&nbsp;" + data.string;
```

I don't see any difference when the mesh is toggled to be `visible=false` or `visible=true`.

`opacity` doesn't seem to do anything either, for instance with the change

```javascript
nv1.setMeshProperty(nv1.meshes[0].id, 'opacity', checkbox1.checked ? 1.0 : 0.0);
```

I still don't see a difference.

hanayik:

- closes #762


neurolabusc:
@hanayik this nicely solves the problem. Can I suggest that the demonstration web page and test use a scene with two meshes, where one can selectively hide or show just one of the meshes. I would suggest [this demo](https://niivue.github.io/niivue/features/meshes.html).

neurolabusc:
Looks good to me.

PR #779, opened 2023-12-05
Test playwright conversion 2

cdrake:
Converts the following tests to Playwright (with snapshots)

- test.clipPlaneAxial.js
- test.clipPlaneCoronal.js
- test.clipPlaneSagittal.js
- test.colorMaps.js
- test.connectomeLoadLegacy.js
- test.defaultOptions.js
- test.fileFormats.js
- test.frac2mm.js
- test.frac2vox.js
- test.frame4DValueOnLocationChange.js


hanayik:
@cdrake , these snapshots all have `linux` in the file name. Is it possible to remove that so that the snapshot names are not platform dependent? Ideally, we want the snapshots to be identical across platforms (like they were with Jest).

hanayik (approved):

PR #787, opened 2023-12-07
TypeScript Migration - NiiVue

jens-ox:
This is the final TypeScript migration PR. With this PR, all JavaScript code is migrated to TypeScript. 

As follow-ups, the following things can be done:
- I added many TODOs and some FIXMEs in the code. The FIXMEs should be addressed soon.
- use more explicit number array types where possible (e.g. `vec3` instead of `number[]`)
- use getters/setters to centrally throw on `undefined` or `null` (e.g. `this.canvas` in `NiiVue`)
- Add Vitest unit tests and code coverage
- Mark internal methods and fields as `private`

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -33,9 +33,14 @@ module.exports = {
   // separate ESLint rules for TS files for now
   overrides: [
     {
-      files: ['**/*.ts'],
+      files: ['src/**/*.ts'],
+      parserOptions: {
+        project: ['./tsconfig.json']
+      },
       rules: {
         '@typescript-eslint/explicit-function-return-type': 'error',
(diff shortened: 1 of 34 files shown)


neurolabusc:
@hanayik this PR should close the following issues: #804 #799 #797 #796 #795 #793 #792

hanayik:
> As with any refactoring undertaking of this size, I can't promise that there are no errors in it, but the tests work fine and I've manually compared all demos to the state before any files were migrated to TypeScript, and they look the same.

Amazing work 😊. I think it would be good to add more tests for some of the cases that had issues during the migration. But that can be part of the next phase of adding in vitest (and the puppeteer --> playwright switch). I think I'll have a go at vitest to try some of that out soon.

hanayik (approved):

Issue #788, opened 2023-12-07
crypto API is not available over HTTP


jennydaman:
https://github.com/niivue/niivue/commit/e0a305a341f261cd1f3184bea50249d973d741e1#diff-55d3a3409e9d3494318a4152a3a0306544efa3229fa33161e4d5e5a0917828c8R113

In e0a305a341f261cd1f3184bea50249d973d741e1 usage of `uuid` was replaced by the `crypto` module, which is (mostly) unavailable when a site is served over HTTP. This is inconvenient for our use cases in our hospital's intranet, could we use a substitute instead? e.g. the [`crypto.getRandomValues`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues) function is (the only one) available.

PR #789, opened 2023-12-07
Fix .js importants and replace crypto.randomUUID() 

jennydaman:

- #788

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,6 +24,7 @@
         "@types/array-equal": "^1.0.2",
         "@types/node": "^20.9.5",
         "@types/ungap__structured-clone": "^0.3.3",
+        "@types/uuid": "^9.0.7",
         "@typescript-eslint/eslint-plugin": "^6.12.0",
         "@typescript-eslint/parser": "^6.12.0",
         "clean-jsdoc-theme": "^4.2.6",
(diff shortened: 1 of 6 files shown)


hanayik (approved):

hanayik:
@jennydaman , can you add types for uuid? 

https://www.npmjs.com/package/@types/uuid

hanayik:
Ok, rebase on main and let’s see how it goes.

jennydaman:
@hanayik bundling with `tsup` is not working as it should, I might need your help to figure this one out.

The `uuid` package has different modules for browser and node, and the browser-specific files to use are set in its `package.json`.

https://github.com/uuidjs/uuid/blob/bc46e198ab06311a9d82d3c9c6222062dd27f760/package.json#L41-L47

The `browser` section of `package.json` is supposed to be read by the bundler. In our case, the bundler is `tsup` (which itself uses `esbuild`).

Running `npx tsup` creates the file `dist/index.js` as expected, however the node files are bundled instead of the browser versions.

```
$ grep -F '// node_modules/uuid' dist/index.js
// node_modules/uuid/dist/esm-node/rng.js
// node_modules/uuid/dist/esm-node/stringify.js
// node_modules/uuid/dist/esm-node/native.js
// node_modules/uuid/dist/esm-node/v4.js
```

Do you have any ideas for a proper fix? Or we could use a different library, e.g.

- https://github.com/lukeed/uuid#lukeeduuid
- https://github.com/ai/nanoid

jennydaman:
@hanayik The default platform for tsup is "node"

Setting platform to "browser" causes daikon to fail to build: esbuild errors on `lib/charLS-DynamicMemory-browser.js`

```
✘ [ERROR] Could not resolve "fs"

    node_modules/daikon/lib/charLS-DynamicMemory-browser.js:36:735:
      36 │ ...rintErr"])Module["printErr"]=function printErr(x){process["stderr"].write(x+"\n")};var nodeFS=require("fs");var nodePath=require("path");Module["read"]=function read(filename,binary){filename=nodePath["norma...
         ╵                                                                                                          ~~~~

  The package "fs" wasn't found on the file system but is built into node. Are you trying to bundle
  for node? You can use "platform: 'node'" to do that, which will remove this error.

✘ [ERROR] Could not resolve "path"

    node_modules/daikon/lib/charLS-DynamicMemory-browser.js:36:762:
      36 │ ...=function printErr(x){process["stderr"].write(x+"\n")};var nodeFS=require("fs");var nodePath=require("path");Module["read"]=function read(filename,binary){filename=nodePath["normalize"](filename);var ret=nod...
         ╵                                                                                                         ~~~~~~

  The package "path" wasn't found on the file system but is built into node. Are you trying to
  bundle for node? You can use "platform: 'node'" to do that, which will remove this error.

ESM Build failed
```

My understanding is that  `lib/charLS-DynamicMemory-browser.js` was (pre-)compiled with emscripten, which produces a JavaScript file which attempts to be compatible with both node and the browser. However, esbuild is angry about how it tries to support node.

PR #790, opened 2023-12-07
Replace UUID

hanayik:

- #789


hanayik:
@jennydaman , I think the proposed uuid package change is ok. Thanks for sorting it out (and investigating). It might be useful to avoid all uuid dependencies in the future (and have one less external thing that can break). We don't truly need _universal_ uniqueness for IDs in NiiVue, we just used it for simplicity at the time. What matters most is that _within_ a niivue instance each volume or mesh has a unique ID so we can look it up reliably without relying on file names. 

The [algorithm](https://www.rfc-editor.org/rfc/rfc4122.html#section-4.4) seems pretty straightforward, with plenty of examples around to reference.

PR #803, opened 2023-12-10
Typescript fix

neurolabusc:
@jens-ox this refactor removes aliases. It has no functional impact, but does clean up the code.

- Remove uniform location aliases

### DIFF
--- a/demos/features/cactus.html
+++ b/demos/features/cactus.html
@@ -142,7 +142,7 @@
     <main id="container">
       <canvas id="gl1"></canvas>
     </main>
-    <footer id="intensity">&nbsp</footer>
+    <footer id="intensity">&nbsp;</footer>
   </body>
 </html>
 <script type="module" async>
--- a/demos/features/freesurfer.html
+++ b/demos/features/freesurfer.html
@@ -123,7 +123,7 @@
     <main id="container">
       <canvas id="gl1"></canvas>
     </main>
-    <footer id="intensity">&nbsp</footer>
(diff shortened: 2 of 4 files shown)


neurolabusc:
@jens-ox I am not sure, but I think both refreshLayers() and gradientGL() should be `async` and called with `await`.  The gradientGL() and its parent refreshLayers() both create and transform large 3D bitmaps on the GPU. WebGL limits us to 8 active textures at any one time, so the compute shaders of these functions must temporarily reassign texture mappings. I do suspect this may explain some issues with our testing as well as ipyniivue. 

Since refreshLayers is computationally expensive, it is only invoked when changes are applied, and not with each refresh. Therefore, awaiting outcome should not impact interactive performance.

Feel free to implement this now if you agree.

jens-ox (approved):

jens-ox:
@neurolabusc

> I am not sure, but I think both refreshLayers() and gradientGL() should be async and called with await. The gradientGL() and its parent refreshLayers() both create and transform large 3D bitmaps on the GPU. WebGL limits us to 8 active textures at any one time, so the compute shaders of these functions must temporarily reassign texture mappings. I do suspect this may explain some issues with our testing as well as ipyniivue.
> 
> Since refreshLayers is computationally expensive, it is only invoked when changes are applied, and not with each refresh. Therefore, awaiting outcome should not impact interactive performance.

As far as I can tell, both `refreshLayers` and `gradientGL` don't call any async or callback-based APIs, therefore the current state should be identical to marking them as async and awaiting them.

PR #810, opened 2023-12-13
Add vitest and starter unit tests

hanayik:
This PR adds vitest and shows a couple of example unit tests to get things started. We can build this up to cover as much as possible now. `happy-dom` is used to mock DOM elements during testing (like `window`). 

to run the unit tests, type `npm run test:vitest` in your terminal from the root niivue directory.


jens-ox (approved):
LGTM! I would suggest to directly include it in the testing pipeline.

hanayik:
@jens-ox , thanks, and done! I think it makes more sense to call it `test:unit` as well, so have renamed that script. Also excluded files and folders that were not needed for coverage. 

If you have experience with mocking things in vitest, it would be good to see how you construct the test for something like `keyDownListener`

hanayik:
@jens-ox , @neurolabusc , and @mrneont , I have added UMD builds back in as part of this PR. This addresses #816. I know that we removed the UMD builds in favour of an ESM-only option, but downstream projects (AFNI) rely on the UMD build and therefore we should continue to support it. @mrneont, I know this may have interrupted your builds, so apologies for that. 

At this stage, AFNI is essentially the main consumer of the UMD build of niivue since many other projects use a different build and deploy cycle (using modern Javascript ecosystem tooling) compared to AFNI's pure python approach. The main deliverable for niivue is unchanged in this commit, so existing projects that have already adopted Niivue since its port to Typescript (such as niivue-react) will not be affected. The UMD is provided for AFNI, and all other consumers of niivue will get the ESM or handle their own module bundling.  

Let me know if there is anything controversial here, @jens-ox since you did a lot of work on the recent tooling up-fit.

@mrneont , I have tested the latest UMD build produced by this branch in your APQC demos that were shared with me. This works as expected in the limited testing.

jens-ox:
@hanayik this is fine for me. Is there any particular reason for AFNI not being able to consume ES Modules? I.e. might adding `type="module"` [here](https://github.com/afni/afni/blob/d2ff26cd902fbd56327a8b0b67218246875fba0a/src/python_scripts/afnipy/lib_apqc_niivue.py#L368) be enough?

mrneont:
@jens-ox : We aren't unable to consume ES files, it was just from the beginning we have been using UMD ones, dating back when there didn't seem to be a major reason to use one or the other.  Since August, people have been baking the latter into their HTMLs, so we will still have to be able to support those, while we do plan to move future work to the new ES ones.  We only just found out about the UMD no longer planning to be supported on Friday, and having a bit of an adjustment/transition overlap will help us.

@hanayik : I appreciate the continuation of UMD, that will help us in this transition period.  I anticipate moving to using ES from the new year, but as we discussed, it will take a bit of recoding because of the ordering of code blocks within the mini-HTML files that get injected.  I imagine that will just take a bit of tinkering but not be too bad.

Thanks for helping with this.  Everyone who has seen the NiiVue functionality in the APQC HTMLs has really enjoyed it and viewed it as helpful (including me, of course!).

Issue #769, opened 2023-11-30
TypeScript migration - follow-ups


jens-ox:
I would recommend doing the following things after the TypeScript migration (see #757, #669, #746):

- **Unit Testing**
Currently, all tests run via a browser instance, either using Puppeteer or using Playwright (only Playwright in the future). For tests that don't rely on rendering things (e.g. testing loaders and transformations), unit tests will be faster and will allow things like checking code coverage. I would recommend using [Vitest](https://vitest.dev/) for that.
- **Methods with many parameters: use parameter objects**
Some methods inside NiiVue have tons of parameters. This means that users have to pay attention to the ordering of the parameters and makes optional parameters harder, as they have to be at the end of the parameter list. Instead, I would recommend using parameter objects (like `loadFromFile` in `NVMesh`) for methods with more than 3 parameters.
- **Logger: Use standardised logging library**
~Instead of shipping a custom logger (that is currently only sparsely used), use an established logging library like [Pino](https://github.com/pinojs/pino).~
@hanayik: pino didn't play nice with the build system. Opted for a homebrewed simple logger that replicated what pino was providing in the browser. 
- **Class methods/fields: use visibility modifiers**
By default, all class fields and methods are public. To make the interfaces easier to understand for users, fields and methods should be annotated with [visibility modifiers](https://www.typescriptlang.org/docs/handbook/2/classes.html#member-visibility).

hanayik:
> I would recommend doing the following things after the TypeScript migration (see #757, #669, #746):
> 
> * **Unit Testing**
>   Currently, all tests run via a browser instance, either using Puppeteer or using Playwright (only Playwright in the future). For tests that don't rely on rendering things (e.g. testing loaders and transformations), unit tests will be faster and will allow things like checking code coverage. I would recommend using [Vitest](https://vitest.dev/) for that.

I agree, I was actually looking into this recently and used [threejs](https://github.com/mrdoob/three.js) as a template for what this could look like. `vitest` looks promising for this.

PR #822, opened 2024-01-11
remove console logs; update logger to pino

hanayik:

- add pino logger (see #769)
- add new eslint rule to avoid `console.log` statements making it into builds

new `logLevel` can be set to `debug | info | warn | error | silent` (default is `info`)

I have tried to do this in a non-invasive way, meaning that I wrapped the `pino` logger using our existing logger in order to minimise code changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,7 +39,6 @@
         "jest": "^29.5.0",
         "jest-image-snapshot": "^6.1.0",
         "jest-puppeteer": "^8.0.5",
-        "pino": "^8.17.2",
         "prettier": "^3.1.0",
         "puppeteer": "^19.7.5",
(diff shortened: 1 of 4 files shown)


hanayik:
@jens-ox and @neurolabusc , the `pino` logger was actually causing build and test errors since I couldn't get it to play nice with `tsup/esbuild`, but worked fine in vite's dev mode. 

I chose to remove pino and implemented a simple logger that replicates the basic functionality that pino was providing us. I have also added unit tests for the logger to ensure that it works (with logLevels) as it should. I think the logging in niivue is much improved now.

I have removed all `console.log` statements and everything that is logged is done via the new logger wrapper.

cdrake (approved):
LGTM

PR #843, opened 2024-01-18
update saveImage logic

### LINKED ISSUE #836: Retrieving NIfTI File After Drawing
My requirement is to send the NIfTI file as a response back to the API after a drawing operation, but I am facing difficulty in finding the option to retrieve the volume/NIfTI file post-drawing. I dont want to download  instead I want to get that niftii.

If there are specific methods or functions that handle the retrieval of NIfTI files, I would be grateful.

hanayik:

- updates saveImage logic to fix #836 

saveImage now returns the Typed image array if the filename is empty. The arguments have also been converted to an object with default values.


neurolabusc:
@hanayik this looks great. It might be worth documenting that this function returns the entire NIfTI file (header and image data). The header is the first 352 bytes, and the image data is dim1 * dim2 * dim3 * nbyper. So for the test case with the mni152 image we have 352 + (207 * 256 * 215 * 1).

hanayik:
@neurolabusc and @cdrake , I have bumped the minor version number in this PR since the `saveImage` method call has been changed.

PR #845, opened 2024-01-18
Accelerate TRK and TCK loading

neurolabusc:

- This [doubles the speed](https://github.com/neurolabusc/trx-javascript/tree/main) for loading TRK and TCK format tractography data
- @jens-ox the new javascript code runs, but the TypeScript generates errors as the functions return vertices are of type `Float32Array` rather than `number[]` and offsets of of type `Uint32Array` rather than `number[]`. My own sense is that **all** tractography data types (TRACT, TT, TRX, TRK, TCK, and VTK) should return these typed arrays. Do you have a strong opinion on this?

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -104,7 +104,6 @@
       function doReset() {
         nv1.setDefaults(defaults, true);
       }
-      await nv1.loadMeshes([{ url: "./ILF_L.tt.gz"}]);
     </script>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


jens-ox:
@neurolabusc using typed arrays everywhere makes sense to me. Do you want me to change the code accordingly?

neurolabusc:
@jens-ox I would be grateful if you could change these to be always typed arrays.

jens-ox:
#847

hanayik (approved):

PR #854, opened 2024-01-23
Refactor niivue functions to child objects

cdrake:
Moved function implementations with no member data access to NVUtilities(sph2cartDeg and vox2mm) and NVImage specific functions to NVImage (vox2frac, frac2vox, mm2frac, and frac2mm) to NVImage. Left function stubs in niivue. 

Converted the following jest tests to vitest unit tests:
- test.sphere2cart.js
- test.vox2frac.js
- test.vox2mm.js
- test.zerosLike.js

### DIFF
--- a/src/niivue/index.ts
+++ b/src/niivue/index.ts
@@ -8633,7 +8633,7 @@ export class Niivue {
     if (this.volumes.length <= volIdx) {
       return [0, 0, 0]
     }
-    
+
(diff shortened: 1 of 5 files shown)


hanayik (approved):
@cdrake , looks good to me. I think we should remove the "not included in public docs" comments in the future, but that can be a general cleanup step that we do later on. We can now exclude this by setting methods as private as needed (if they don't need to be in the developer docs).

PR #859, opened 2024-01-24
Fix nullable parameter type of setModulationImage

jennydaman:


jennydaman:
Seems like the problem runs deeper, I will investigate.

hanayik:
@jennydaman, are you still looking into this or should someone else take it? Happy to, just didn't want to if you were still investigating. 

LMK.

jennydaman:
@hanayik would you take over?

PR #866, opened 2024-02-06
Test ts conversion add snapshot delay

cdrake:
Converted javascript tests to typescript. Added 1 sec delay before snapshots.


hanayik:
@cdrake , there was a screenshot comparison failure here: https://github.com/niivue/niivue/actions/runs/7795564460/job/21258713607#step:6:79

Could you investigate? Also, can you make sure the tests exit (fail) if there is a single failed test?

PR #886, opened 2024-02-15
Test conversion ts

### LINKED ISSUE #735: add test for thumbnail resizing
- take one snap shot on first page load
- then resize the window (programmatically) and take another.

### LINKED ISSUE #824: enhance unit testing (remove non-rendering tests from puppeteer/playwright)
originally from #769 

### Unit Testing
Currently, all tests run via a browser instance, either using Puppeteer or using Playwright (only Playwright in the future). For tests that don't rely on rendering things (e.g. testing loaders and transformations), unit tests will be faster and will allow things like checking code coverage. I would recommend using [Vitest](https://vitest.dev/) for that.

### LINKED ISSUE #882: dist of NiiVue contains incorrect browser polyfills
### Context

We are tracing a rare bug where the mere inclusion of NiiVue into a project already using hammer.js and cornerstone-tools breaks cornerstone.

### Bug

The build of `@niivue/niivue` adds polyfills to the `dist/index.js` file. These polyfills are both activated incorrectly and also implemented incorrectly. I have `@niivue/niivue@0.39.0` installed using `npm`. Here are the contents of `node_modules/@niivue/niivue/dist/index.js` lines 17562 to 17567:

```javascript
      if (!Array.forEach) {
        Array.prototype.forEach = function(fn) {
          for (var i2 = 0; i2 < this.length; i2++)
            fn(this[i2], i2, this);
        };
      }
```

1. The expression `!Array.forEach` does not correctly check for whether the polyfill is necessary.

image

2. The function `Array.prototype.forEach` is 2-arity: the second parameter is an option parameter used as the value for `this`. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#thisarg

```javascript
// ECMAScript 2009 class definition
function Foo() {
  this.hello = { value: "hello" }
}

Foo.prototype = {
  printHello: function() {
    console.log(this.hello.value)
  }
}

var foo = new Foo();

[foo].forEach(Foo.prototype.printHello, foo)  // prints "hello"

// poisonous polyfill
Array.prototype.forEach = function(fn) {
  for (var i2 = 0; i2 < this.length; i2++)
    fn(this[i2], i2, this);
};

[foo].forEach(Foo.prototype.printHello, foo)  // Uncaught TypeError: this.hello is undefined
```

cdrake:

- Converts test from JavaScript to TypeScript (via @cdrake)
- closes #735 (via @hanayik)
- closes #882 (via @hanayik)
- adds piped stdout and stderr to playwright console output from browser window
- closes #824 until more vitest unit tests added later


hanayik:
@cdrake , I have added a couple of commits to this PR. They should fix #882. I updated the builds to use a separate tsup config that targets a build for our tests, and another tsup config that is aimed at building a leaner niivue javascript bundle. 

@jens-ox, and @jennydaman, it might be nice for the two of you to take a look at these specific [changes](https://github.com/niivue/niivue/pull/886/commits/1c056b084472f8bfa60fa1c568eb5422bbef814b) too.

cdrake:
Sounds good to me.  Let me know if you would like me to make any other changes.

cdrake:
Looks like testing is broken. I will try to get testing working again.

hanayik:
@cdrake, please review my latest commit (fix pretest). I removed all the *.js output files that are created by `tsc`. I also updated a screenshot that was wrong. I created a nodejs script to copy files (so it can make the output folder if needed). I also updated gitignore.

hanayik (approved):

PR #894, opened 2024-02-22
[FIX] eslint and minor bug

dlevitas:

- Fix eslint issues and minor bug in `concat_json2trx.mjs`

### DIFF
--- a/src/concat_json2trx.mjs
+++ b/src/concat_json2trx.mjs
@@ -1,14 +1,14 @@
-//Concatenate all tck files in a folder to a single trx file
+// Concatenate all tck files in a folder to a single trx file
 // each tck file is saved as a unique group
-//Usage:
+// Usage:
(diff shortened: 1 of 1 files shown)


hanayik:
@dlevitas , I'm merging this in since it is NOT into the main branch. However, I don't think `concat_json2trx.mjs` should be in the repo at all. It is not intended to be used in the browser, and looks to be more of a convenience script rather than something that should be in `/src`

neurolabusc:
@hanayik thanks. This  fork is not meant to be merged into the main NiIVue branch anytime soon. Rather, this fork will explore how brainlife can use NiiVue for QA. This will go both directions: move brainlife from proprietary formats (e.g. json files for streamlines) to community formats (trx) and then make sure NiiVue can support all the brainlife features (e.g. NiiVue supports TRX groups, but does not yet support TRX per-group colors). So this branch allows us to meet half way.

PR #899, opened 2024-02-29
Fix canvas parent resizing issue

### LINKED ISSUE #861: [BUG] with isResizeCanvas: canvas does not resize when parent is resized
Automatic resizing of the canvas with the option `isResizeCanvas` responds to when the window is resized, _not_ when the canvas' parent element is resized.

Copy-paste this example code into `src/index.html`, then run `npm run dev`

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
    <meta name="viewport" content="width=device-width,initial-scale=1.0" />
    <title>NiiVue</title>

    <style>
      html, body, main, #sidebar-container, #niivue-container {
        height: 100%;
        width: 100%;
      }

      #sidebar-container {
        display: flex;
      }

      #gl1 {
        height: 100%;
        width: 100%;
      }

      #sidebar {
        width: 250px;
        transition: width 1s linear;
      }

      #sidebar.expanded {
        width: 800px;
      }
    </style>
  </head>
  <body>
    <noscript>
      <strong>niivue requires JavaScript.</strong>
    </noscript>
    <div id="sidebar-container">
      <div id="sidebar">
        I am the sidebar.
        <br />
        <label for="sidebar-checkbox">Toggle sidebar size</label>
        <input type="checkbox" id="sidebar-checkbox" />
      </div>
      <div id="niivue-container">
        <canvas id="gl1"></canvas>
      </div>
    </div>
    <script type="module" async>
      import { Niivue } from "./niivue/index.ts";
      var nv1 = new Niivue({isColorbar: true});
      nv1.attachTo("gl1");
      await nv1.loadVolumes([{ url: "../demos/images/pcasl.nii.gz", colorbarVisible: true }]);

      document.getElementById("sidebar-checkbox").onclick = function() {
        if (this.checked) {
          document.getElementById("sidebar").classList.add('expanded');
        } else {
          document.getElementById("sidebar").classList.remove('expanded');
        }
      }
    </script>
  </body>
</html>
```

When the sidebar is expanded, Niivue canvas is vertically "smushed" and where I click on the canvas is offset. The bug fixes itself if the window is resized (which calls `nv.resizeListener()`). A work-around would be to call `nv.resizeListener()` in the `onclick` callback, however this can be complicated to do in real-world projects (where JS/CSS frameworks are in use).

hanayik:
- closes #861 

@jennydaman , please checkout the `bug-isresizecanvas` branch and then run `npm run dev`. The example web page that loads should automatically resize the parent element of the canvas, and trigger a proper redrawing of the canvas. 

So, now niivue will be aware of window resize events AND the resize events of the `canvas` parent element. This implementation uses the `ResizeObserver` which should be available in all major browsers. 

@jennydaman , you should be able to remove your [workaround](https://github.com/jennydaman/ChRIS_ui/commit/8b1e3ac96629967712e6370e074d897f0b8f32b0) once this is published in a new version of Niivue.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -57,7 +57,7 @@
       />
       <button id="about">About</button>
     </header>
-    <main>
+    <main id="canvas-container">
       <canvas id="gl1"></canvas>
     </main>
     <footer>
@@ -118,6 +118,17 @@
         };
         cmapEl.appendChild(btn);
       }
+      // TEMPORARY! TODO: remove before merging into main
+      // every 500ms make the canvas container 2px shorter and narrower
+      let resizeInterval = setInterval(() => {
+        let canvasContainer = document.getElementById("canvas-container");
+        canvasContainer.style.width = canvasContainer.offsetWidth - 2 + "px";
+        canvasContainer.style.height = canvasContainer.offsetHeight - 2 + "px";
+        // stop when the canvas container is 500px wide
+        if (canvasContainer.offsetWidth <= 500) {
+          clearInterval( resizeInterval);
+        }
+      }, 16);
     </script>
   </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -118,17 +118,6 @@
         };
         cmapEl.appendChild(btn);
       }
-      // TEMPORARY! TODO: remove before merging into main
-      // every 500ms make the canvas container 2px shorter and narrower
-      let resizeInterval = setInterval(() => {
-        let canvasContainer = document.getElementById("canvas-container");
-        canvasContainer.style.width = canvasContainer.offsetWidth - 2 + "px";
-        canvasContainer.style.height = canvasContainer.offsetHeight - 2 + "px";
-        // stop when the canvas container is 500px wide
-        if (canvasContainer.offsetWidth <= 500) {
-          clearInterval( resizeInterval);
-        }
-      }, 16);
     </script>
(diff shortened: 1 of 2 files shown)


jennydaman:
@hanayik this is what I see: the resizing works, however there is a new bug where the canvas shrinks to whatever size it wants to be, rather than obeying its CSS rules which want it to occupy 100% of parent.

The new bug is present in both Chromium and Firefox.

(apologies for the screen flickering in my screencast.)

https://github.com/niivue/niivue/assets/20404439/a0e845ab-111c-49fd-aec5-751dad3db05a

hanayik:
> @hanayik this is what I see: the resizing works, however there is a new bug where the canvas shrinks to whatever size it wants to be, rather than obeying its CSS rules which want it to occupy 100% of parent.
> 
> The new bug is present in both Chromium and Firefox.

@jennydaman are you referring to the shrinking animation? If so, that is intentional to show that the resizing works. It is just temporary to illustrate the effect. You can verify in src/index.html.

jennydaman:
Ah, I see your `// TEMPORARY!`. In that case, works great! I tried on Chromium, Firefox, and iPhone Firefox (which is reskinned webkit).

PR #904, opened 2024-03-04
Fix getFrame4D method to use correct property

### LINKED ISSUE #902: [Bug] with getFrame4D: returns nFrame4D instead of frame4D
The function Niivue.getFrame4D() returns nFrame4D instead of frame4D in the source code, which seems inconsistent with its description.

hanayik:

- closes #902 

`getFrame4D` was using the `nFrame4D` property which is wrong. `nFrame4D` was intended to represent the max number of volumes. 

Upon investigation, I discovered that we use both `nFrame4D` and `nTotalFrame4D` which seems redundant. 

@neurolabusc , are you in favour of removing the `nTotalFrame4D` property from `NVImage`? I can do so if you agree.

### DIFF
--- a/src/niivue/index.ts
+++ b/src/niivue/index.ts
@@ -6025,7 +6025,7 @@ export class Niivue {
    */
   getFrame4D(id: string): number {
     const idx = this.getVolumeIndexByID(id)
-    return this.volumes[idx].nFrame4D!
+    return this.volumes[idx].frame4D!
   }
 
   // not included in public docs


neurolabusc:
@hanayik these are not redundant. You can see the difference in the [live demo](https://niivue.github.io/niivue/features/timeseries2.html). The `nFrame4D` reports the number of currently downloaded volumes, while `nTotalFrame4D` reports the total number of volumes. Consider the live demo, where the volume is loaded with `limitFrames4D: 5`. In this example, the `nFrame4D` is 5, but `nTotalFrame4D` reveals that the series has 10 volumes. The live demo is interactive: click on the ellipsis (`...`) in the graph to load all the remaining volumes, at with case `nFrame4D` will be reported as 10.

Issue #920, opened 2024-03-21
Refactor texture slots


neurolabusc:
Our current developer [documentation](https://github.com/niivue/niivue/blob/main/docs/development-notes/webgl.md) suggests WebGL requires a minimum of [8 MAX_COMBINED_TEXTURE_IMAGE_UNITS](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/activeTexture). However, NiiVue is based on WebGL2 which has a [minimum of 32 total loaded textures, with a fragment program able to access at least 16 concurrently](https://webgl2fundamentals.org/webgl/lessons/webgl-cross-platform-issues.html#device-limits). The upcoming pull request makes two changes:
 - Uses named constants rather than generic WebGL constants. So instead of referring to `gl.TEXTURE7` the code refers to `TEXTURE7_DRAW` so it is clear that the drawing texture is being made the active texture for slot 7.
 - Update the [documentation](https://github.com/niivue/niivue/blob/main/docs/development-notes/webgl.md) to reflect WebGL2 capabilities and describe the named constants.

PR #921, opened 2024-03-21
WebGL Texture refactor (https://github.com/niivue/niivue/issues/920)

neurolabusc:

- #920

### DIFF
--- a/docs/development-notes/webgl.md
+++ b/docs/development-notes/webgl.md
@@ -10,16 +10,24 @@ WebGL is a modern shader-based language. It lacks the fixed function pipeline, q
 
 ##### Textures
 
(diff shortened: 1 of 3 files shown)


docs/development-notes/webgl.md
cdrake, line 1: It makes sense and gives a good explanation about why we have assigned texture slots as either reserved or transient.  I think it might be good for consistency to label texture slots 0 - 7 as reserved/persistent.

src/shader-srcs.ts
cdrake, line 1: I think it's an improvement to change the sampler names. It makes following the shader code easier.

cdrake (approved):
I think this PR helps make our code easier to follow.

Issue #935, opened 2024-04-01
Connectome refactor and consistency


neurolabusc:
There is an opportunity to refactor connectome support. At the moment, **both** nvmesh.ts and nvconnectome.ts include the functions `updateConnectome()` and redundant code for reading tabular connectome files. The nvmesh code is used for reading files and only supports tabular files. While many of our tests and live demos (like [this one](https://niivue.github.io/niivue/features/shiny.volumes.html)) use `loadMeshes()` to load tabular connectomes, the same call fails if provided with a sparse connectome. The commit removes the `loadConnectomeFromJSON()` and `updateConnectome()` functions from nvmesh and ensures that loadMeshes handles any form of connectome.

Not handled in this commit, I would suggest that `loadConnectomeFromFreeSurfer()` be moved from  nvmesh.ts to nvconnectome.ts.

Issue #934, opened 2024-03-31
Multiuser basic 3D live demo regression


neurolabusc:
The [Multiuser basic 3D](https://niivue.github.io/niivue/features/multiuser.3d.html) is broken. It attempts to import NiiVue, but the script is not a module.

Issue #933, opened 2024-03-31
Multiuser colormaps live demo regression


neurolabusc:
The [Multiuser colormaps](https://niivue.github.io/niivue/features/multiuser.imageoptions.html) live demo does not display any colormaps. I think this regression occurred when the script became a module.

Issue #932, opened 2024-03-31
AFNI volumes animation


neurolabusc:
The animate button on the [AFNI volumes live demo](https://niivue.github.io/niivue/features/afni.html) is no longer functional. This regression was an unintended issue of [issue 902](https://github.com/niivue/niivue/issues/902). The live demo should use `nFrame4D` not `getFrame4D`.

neurolabusc:
The issue902 regression also impacts the [multiuser time series](http://192.168.0.151:8080/features/multiuser.timeseries.html), [4D with initial thumbnail](https://niivue.github.io/niivue/features/timeseries.html), and [4D with partial load](https://niivue.github.io/niivue/features/timeseries2.html) live demos

PR #936, opened 2024-04-01
Lots of little fixes

neurolabusc:

- #935
- #934
- #933
- #932 
- #931 
- #930 
- #929 
- #928 
- #927 
- #926 
- #774


src/niivue/index.ts
```diff
@@ -104,6 +104,7 @@ export * from '../nvdocument.js'
 export { NVUtilities } from '../nvutilities.js'
 export { LabelTextAlignment, LabelLineTerminator, NVLabel3DStyle, NVLabel3D } from '../nvlabel.js'
 export { NVMeshLoaders } from '../nvmesh-loaders.js'
+export { Connectome, LegacyConnectome} from "../types.js"
```
neurolabusc: @hanayik adding this line causes `npm run dev` to fail for me. I think it should be removed from this PR.

hanayik: The line just needs to be modified, rather than removed. I will submit a new commit once local tests pass.

hanayik (approved):

PR #865, opened 2024-02-04
Clean up build setup

jens-ox:
@neurolabusc reached out to me due to working with `npm run demo` being really slow and a non-pleasant setup in general, as restarting it was required on changes.

This PR replaces `npm run demo` with a structurally similar setup to `npm run dev` that uses Vite to serve the demos. To build the demos, use `npm run demo:build`. To access the build demos like before, use `npm run demo:serve`.

A few questions and points:
- As this change will certainly need some discussion, I only adapted `features/basic.multiplanar.html` so far.
- This PR will require changes to how the demos are hosted on GitHub Pages. This is not a complicated change (use [`actions/deploy-pages`](https://github.com/actions/deploy-pages) on push to main) and I can take care of it within this PR once no other discussion points are open.

Changes:
- prepare some HTML files to be actually valid HTML (i.e. no non-tags in text like `<lp_sigma>`)
- remove unused `vite.config_inject.ts`
- add new `vite.config.demos.ts`
- adapt NiiVue import in demo HTMLs to point to source code
- `npm run demo`: runs the demos in dev mode with HMR and all other bells and whistles, startup and iterations should be near instantaneous.
- `npm run demo:serve`: runs a built version of the demos, similar to how it worked before.
- `npm run demo:build`: invoked by `npm run demo:serve` (and by the deploy step later) to build the demos.

PR #937, opened 2024-04-03
Package maintenance

jens-ox:
@neurolabusc contacted me regarding performance optimizations using typed arrays. In order to facilitate doing performance optimization, this PR addresses a couple of points:

- update all dependencies

**Upcoming:**

- separation of Prettier and ESLint for better linting performance (see [this blog post](https://www.joshuakgoldberg.com/blog/you-probably-dont-need-eslint-config-prettier-or-eslint-plugin-prettier/))
- force parameter objects for functions with more than 3 parameters
- reduce usage of `NaN` and `null` in favor of `undefined` (most of them were caused by having to pass `null` values so that later parameters in method invocations could be passed)
- replace `number[]` with better fitting array types like `UInt8Array`, `vec4` etc

**Help needed:**

Some methods are making the switch to typed arrays really hard, e.g. `makeColoredCylinder` and `makeColoredSqphere`, as they modify the length of arrays passed as params (`colors.push(...clrs)`). The length of the modified array needs to be known at instantiation of the source array. Alternatively, the methods could return the resulting array, so that the source array can be replaced.

I'll wait with larger code changes until #936 is merged. I will also revive #865 after this PR.


hanayik:
@jens-ox , thanks! I'm working on fixing some issues in #936 today and will hopefully have that merged later.

jens-ox:
@hanayik sounds good! I'm a bit worried about the failing tests here. Not really sure why they're failing.

neurolabusc:
@jens-ox thanks for taking the initiative.

We can certainly help optimize this. While .push is convenient and makes for readable code, it is much slower that using preallocated arrays. I think your solution of having `makeColoredCylinder` and `makeColoredSphere` returning just the new vertices/indices that can be inserted at the correct location in an existing pre-allocated array is the way to go. 

Its worth noting that I already ported `linesToCylinders()` from using .push() to  pre-allocating (`f32 = new Float32Array(npt * f32PerVertex` and `idxs = new Int32Array(nidx)`) and had big performance gains.

hanayik:
> @hanayik sounds good! I'm a bit worried about the failing tests here. Not really sure why they're failing.

@jens-ox , I'll look into it, I _believe_ the puppeteer tests are redundant now that the playwright tests are finalised. It may be useful to disable them soon :). The unit tests will need to be separated from the puppeteer invocation as well.

hanayik:
@jens-ox, I have refactored the linting and testing github actions and removed the redundant puppeteer tests in #936. There is also _local_ support for testing snapshots from demo pages (`npm run test-demos`), but this is not implemented in github actions yet.

Could you rebase your branch? There may be a few merge conflicts in this one that we'll need to sort through.

hanayik:
@jens-ox , I have merged #943.

hanayik (approved):
@jens-ox , approving and merging this now. The changes to `tests/` won't take any effect though since puppeteer is no longer used. These tests should be removed in a future PR. 

Our tests can be cleaned up and reorganised now that we only use `vitest` and `playwright`.

PR #942, opened 2024-04-04
only allow idModulation to be a string

hanayik:
Previously, the `idModulation` argument to `setModulationImage` could accept a string or `null`. I have updated `setModulationImage` to only accept a string (or empty string) rather than both strings and `null`. The demos that use this feature have also been updated. To clear the modulation, a user must supply an empty string as the value for `idModulation`. 

I have checked that the AFNI usage of this feature is not impacted. 

- closes #859


neurolabusc:
Looks good to me.

PR #943, opened 2024-04-04
Optimize mesh types

neurolabusc:
@jens-ox this PR uses typed arrays for WebGL meshes, as you note in your [PR 937](https://github.com/niivue/niivue/pull/937). The number[] arrays are still used for some transient usage (where the larger memory footprint has less impact than long term storage). However, all mesh indices are unified as Uint32 and all mesh vertex positions are unified as Float32. I think this addresses the `help needed` request from your PR.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -99,7 +99,7 @@
       nv1.opts.yoke3Dto2DZoom = true;
       await nv1.loadVolumes(volumeList1);
       var layerList = [{url: "../demos/images/mni152.SLF1_R.tsf"}];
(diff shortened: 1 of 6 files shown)


jens-ox (approved):
This looks very good! I think some instantiations could be saved by moving more methods to typed arrays, but that can be done in the future.
The codebase currently contains typed array instantiation via constructor calls (`new UInt8Array([1,2,3])`) and via `.from` calls (`UInt8Array.from([1,2,3])`). While it's not really necessary, aligning might make sense.

PR #946, opened 2024-04-07
Fix clone method to generate a new ID

### LINKED ISSUE #945: clone() should assign a new ID rather than using the donor image ID

hanayik:
This PR ensures that a new ID is given to cloned images in order to avoid ID based conflicts. 

A new unit test has been added for NVImage as well that will make sure this issue doesn't happen in the future. 

- closes #945


hanayik:
This is a trivial fix, so merging now (without reviewer approval)

PR #959, opened 2024-04-21
better logic for dragMode in touchEnd

hanayik:
add `generateMouseUpCallback` behaviour to `touchEndListener`. This enables the same behaviour that we see in `mouseUpListener` by reusing our existing functions. 

Also updated the handling of different `DRAG_MODE` values in the `touchEndListener` event.

### DIFF
--- a/src/niivue/index.ts
+++ b/src/niivue/index.ts
@@ -1356,12 +1356,21 @@ export class Niivue {
     }
     if (this.uiData.isDragging) {
       this.uiData.isDragging = false
-      // if (this.opts.isDragShowsMeasurementTool) return;
-      if (this.opts.dragMode !== DRAG_MODE.contrast) {
+      // if drag mode is contrast, and the user double taps and drags...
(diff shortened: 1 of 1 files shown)


neurolabusc (changes requested):
Our current behavior for the mouse is that generateMouseUpCallback() occurs at the end of a **mouse drag**, regardless of the `dragMode`. In contrast, your PR will only fire generateMouseUpCallback() at the end of a **touch drag** if the `dragMode` is `callbackOnly`. The callbacks should be generated at the end of **any** drag, with this being the only consequence of if the `dragMode` is `callbackOnly`. I would simplify your code to emulate mouse behavior:

```
      if (this.opts.dragMode === DRAG_MODE.contrast) {
        this.calculateNewRange()
        this.refreshLayers(this.volumes[0], 0)
      }
      this.generateMouseUpCallback(fracStart, fracEnd)
```

neurolabusc (approved):

PR #972, opened 2024-05-14
Fix saving and loading document

cdrake:
- Drag and dropping document was not loading config options
- Loading tractography, and connectomes 
- Saving color map labels


hanayik (changes requested):
@cdrake , there are a few `console.log` lines that look like they were used for debugging. Can you convert these to use the niivue logger instead if they aren't critical logs?

src/index.html
neurolabusc, line 1: @cdrake I only changed this web page to help test your new features. I would not change the current NiiVue ./src/index.html file as it is nice to have a minimal page when using `npm run dev`. Beyond this, this is terrific work.

cdrake: Yes, index.html should be reverted.

cdrake:
> @cdrake , there are a few `console.log` lines that look like they were used for debugging. Can you convert these to use the niivue logger instead if they aren't critical logs?

Console.logs added should be removed or converted to log.error

Issue #976, opened 2024-05-16
Atlas not saving correctly


cdrake:
Atlases (like aal.nii) are not rendered from a saved document

PR #978, opened 2024-05-16
Fix atlas saving/loading and color rendering for colormaplabel

cdrake:
Correctly saves and loads scenes with atlas and other .nii files with voxel offset not equal to 352
Converts color map label lookup table array into Uint8ClampedArray

- [Atlas not saving correctly](https://github.com/niivue/niivue/issues/976) 
- [Color not rendered from volumes with color map label](https://github.com/niivue/niivue/issues/977)
- [Add the capability to save custom data to NVDocument](https://github.com/niivue/niivue/issues/979)

### DIFF
--- a/src/nvimage/index.ts
+++ b/src/nvimage/index.ts
@@ -235,10 +235,7 @@ export class NVImage {
           if (nifti.isCompressed(dataBuffer as ArrayBuffer)) {
             imgRaw = nifti.readImage(this.hdr, nifti.decompress(dataBuffer as ArrayBuffer))
           } else {
-            // imgRaw = nifti.readImage(this.hdr, dataBuffer as ArrayBuffer) // this.hdr.vox_offset
-            console.log('this.hdr.vox_offset', this.hdr.vox_offset)
-            imgRaw = new Uint8Array(dataBuffer.slice(352) as ArrayBuffer)
-            console.log('imgRaw length', imgRaw.byteLength)
+            imgRaw = nifti.readImage(this.hdr, dataBuffer as ArrayBuffer)
           }
         }
         break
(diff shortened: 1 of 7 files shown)


src/nvimage/index.ts
```diff
@@ -3560,7 +3560,7 @@ export class NVImage {
    */
   toUint8Array(drawingBytes: Uint8Array | null = null): Uint8Array {
     const isDrawing = drawingBytes !== null
-    const hdrBytes = hdrToArrayBuffer(this.hdr!, isDrawing)
+    const hdrBytes = hdrToArrayBuffer({ ...this.hdr!, vox_offset: 352 }, isDrawing)
```
hanayik, line 3563: @cdrake and @neurolabusc , by forcing `vox_offset: 352`, is there a possibility of saving a corrupted nifti file if the original file had extensions present? `toUint8Array` could be called on any input image provided to NiiVue, not just images we create with internal functions.

neurolabusc: @hanayik no, when we save an image to NIfTI we remove any extensions added to image, and the image is always at offset 352. This is intentional, as we do only want to retain data we can interpret. @cdrake noted this change is required. Our sample `aal.nii.gz` exhibits this - the header offset is not 352 since it includes atlas labels as text inserted after the binary header and before the image.

cdrake: @hanayik - No, because the toUint8Array method for NVImage (which is what NVDocument uses to serialize volumes) sets the header as 348 bytes (through the hdrToArrayBuffer method), adds 4 bytes and then adds the img array at the end. It does mean we do not preserve extensions using this method.

``` typescript
toUint8Array(drawingBytes: Uint8Array | null = null): Uint8Array {
    const isDrawing = drawingBytes !== null
    const hdrBytes = hdrToArrayBuffer({ ...this.hdr!, vox_offset: 352 }, isDrawing)
    ...
    const img8 = isDrawing ? (drawingBytesToBeConverted as Uint8Array) : new Uint8Array(this.img!.buffer)
    const opad = new Uint8Array(4)
    const odata = new Uint8Array(hdrBytes.length + opad.length + img8.length)
    odata.set(hdrBytes)
    odata.set(opad, hdrBytes.length)
    odata.set(img8, hdrBytes.length + opad.length)
    return odata
  }
```
  
``` typescript
export function hdrToArrayBuffer(hdr: NiftiHeader, isDrawing8 = false): Uint8Array {
  const SHORT_SIZE = 2
  const FLOAT32_SIZE = 4

  const byteArray = new Uint8Array(348)
```

hanayik (approved):

Issue #983, opened 2024-05-17
Support gzipped NVDocuments


cdrake:
Allow users to save and load gzipped NVDocuments

neurolabusc:
@cdrake this should be easy, as we already include [fflate](https://github.com/101arrowz/fflate) with NiiVue. Given the size of scenes, using the asynchronous variants that use web workers seems appropriate. You could even use the same `.nvd` extension when writing either compressed or uncompressed data, and use the [first two bytes](https://github.com/niivue/niivue/blob/d5fd3750d60b8f24f45cd454f4d46bc94ee221a5/src/nvmesh-loaders.ts#L1716) to detect the [magic number 0x1F8B](https://en.wikipedia.org/wiki/Gzipr) as the signature of gzipped data.

cdrake:
@neurolabusc - Will do.  I'm working on it now.

PR #984, opened 2024-05-18
Feature support gzipped doc

cdrake:
Adds support to load gzipped documents (retains .nvd file extension) and by default gzips documents during save.

- [Support gzipped NVDocuments](https://github.com/niivue/niivue/issues/983)


neurolabusc:
@cdrake nice job. This is very elegant. 

I did make one fix that is very unintuitive: streamlines are a very special kind of mesh, using offsets (offset0) and vertices rather than triangle indices (tris). This is overloaded when we create a new mesh, so your documents were storing both offset0[] and tris[] arrays (with copies of the same data). The size penalty is negligible for the toy streamlines in our demos, but could be considerable for large meshes. I have fixed this, so we ALWAYS only save tris[] or offset0[]. My solution is backwards compatible. This reflects an early design decision on my part to overload this function. It is efficient but not intuitive. I am highlighting this fix in the off chance it has an unintended consequence, but it seems to work in my testing.

neurolabusc:
@cdrake I do have two suggestions, though these are not a bugs.

[compressToBase64String()](https://github.com/niivue/niivue/blob/a36b0391320d41de7bb885e1e62fca1c480b3dc0/src/nvutilities.ts#L124) uses compressSync(). On the other hand, [compressStringToArrayBuffer()](https://github.com/niivue/niivue/blob/a36b0391320d41de7bb885e1e62fca1c480b3dc0/src/nvutilities.ts#L133) uses compress(). Given that these functions deal with files of similar size, is there a reason for having one synchronous and the other asynchronous? Stylistically, it is also odd that one function uses `new TextEncoder().encode(input)` while the other uses the fflate strToU8().

I have two minor suggestions that you can feel free to ignore.

First, use strToU8() for both functions simply for consitency:
```
  static async compressStringToArrayBuffer(input: string): Promise<ArrayBuffer> {
    return new Promise((resolve, reject) => {
      compress(strToU8(input), (err, compressed) => {
        if (err) {
          reject(err)
        } else {
          resolve(compressed.buffer)
        }
      })
    })
  }
```
The second suggestion would require making generateLoadDocumentJavaScript() and generateHTML() and any functions that call them (like [here](view-source:https://niivue.github.io/niivue/features/save.html.html)) **async** .  While this might be seen as a breaking change, it is worth noting that this would eliminate warnings like `[Violation] 'click' handler took 1276ms` exhibited by [our live demo](https://niivue.github.io/niivue/features/save.html.html).

I have not implemented all these changes, as you might want to discuss if this is the approach you want to follow. However, I think the core function would look somthing like this:
```
  static async compressToBase64String(string: string): string {
    const buf = strToU8(string)
    return new Promise((resolve, reject) => {
      const uint8Array = new TextEncoder().encode(input)
      compress(uint8Array, (err, compressed) => {
        if (err) {
          reject(err)
        } else {
            resolve(NVUtilities.uint8tob64(compressed))
        }
      })
    })
  }
```

cdrake:
Updated decompressBase64String and compressToBase64String to be async. Updated code to use strFromU8 and strToU8 instead of text encode/decode.  Updated html for download code to use promise

``` typescript
static async decompressBase64String(base64: string): Promise<string> {
    const compressed = atob(base64)
    // convert to an array buffer
    const compressedBuffer = new ArrayBuffer(compressed.length)
    const compressedView = new Uint8Array(compressedBuffer)
    for (let i = 0; i < compressed.length; i++) {
      compressedView[i] = compressed.charCodeAt(i)
    }
    return NVUtilities.decompressArrayBuffer(compressedView)
  }

  static async compressToBase64String(string: string): Promise<string> {
    const buf = await NVUtilities.compressStringToArrayBuffer(string)
    return NVUtilities.uint8tob64(new Uint8Array(buf))
  }
```

``` javascript
 NVUtilities.decompressBase64String(base64).then((jsonText) => {
          var json = JSON.parse(jsonText); // string -> JSON
          var doc = NVDocument.loadFromJSON(json);                
          nv1.loadDocument(doc);
          nv1.updateGLVolume();
        });
```

Issue #999, opened 2024-06-18
Multiple simultaneous clipping planes


hsellik:
Hi

Looking at the source and documentation, there doesn't seem to be a way to clip multiple planes simultaneously. This would allow to build UI sliders for clipping planes as it is done by, for example, [Allen Institute for Cell Science](https://cfe.allencell.org/?cellSelectedFor3D=660632&colorBy=cell-line&dataset=variance_v1&plotByOnX=cell-volume&plotByOnY=dna-volume&selectedPoint%5B0%5D=660632).

What would be the projected effort to build out such functionality?

https://github.com/niivue/niivue/assets/16798824/f5d98d57-1d2a-4a93-a174-d576f194e607

neurolabusc:
I have submitted a new pull request that will create the [clipplanes2 live demo](https://niivue.github.io/niivue/features/clipplanes2.html) when merged. This live demo showcases two new features:
 - nv.setClipPlaneThick(): allows the existing clip plane (which allows oblique cuts) to be shown as a slab with arbitrary thickness.
 - nv.setClipVolume(): defines a clipping volume across columns (x), rows (y) and slices (z).
 
The live demo also shows how they perform differently. The clip plane only influences the background volume and is illustrated with a graphical orientation rectangle [with adjustable color and opacity](https://niivue.github.io/niivue/features/clipplanes.html), while the clip volume influences all volumes. I do like how the Allen Institute Cell Feature Explorer provides two-handled sliders. I use plain HTML for all the live demos to provide minimal, framework agnostic demos. However, for these functions I would recommend using a two-handled slider with your favorite framework (Angular, React, Vue).  


![clip](https://github.com/niivue/niivue/assets/8930807/d3035f94-f961-44f6-afbf-318f5d235d55)

hsellik:
That looks perfect and the assumption about the sliders is very reasonable. 👍 It will be super simple to use when [niivue-react](https://github.com/niivue/niivue-react) gets a prop for this.

neurolabusc:
PR is merged. Please evaluate the [live demo](https://niivue.github.io/niivue/features/clipplanes2.html) and close this issue if it resolves your request.

PR #1000, opened 2024-06-19
Volume clipping

neurolabusc:

- #999 

@cdrake this adds three new properties to the scene data: clipThick, clipVolumeLow, clipVolumeHigh. For backward compatibility it is important that these are set to the default values if not present in a NVD document. Likewise, we should make sure to retain this data when we save scenes as NVD documents.


cdrake (approved):
LGTM - Added a test for clip volume/thickness

Issue #1005, opened 2024-06-24
Improve MRtrix MIF support


neurolabusc:
The upcoming NiiVue commit extends support for the MRtrix MIF format. It supports the compact 1-bit voxels used for masks. To create MIF images of various types, you can follow [Andy Jahn's tutorial](https://andysbrainbook.readthedocs.io/en/latest/MRtrix/MRtrix_Course/MRtrix_01_Download_Install.html). We also want to support the FA and V1 files created by [dwi2tensor](https://mrtrix.readthedocs.io/en/dev/reference/commands/dwi2tensor.html) and [tensor2metric](https://mrtrix.readthedocs.io/en/dev/reference/commands/tensor2metric.html). Note that unlike FSL and DSI-Studio, MRtrix can generate NaN values for V1 vectors.

Once my upcoming patch is provided, a new [live demo](https://niivue.github.io/niivue/features/modulateTrix.html) will visualizing the tensor2metric results as shown below. This commit also includes routines for directly extracting FA and V1 from the MIF files created by dwi2tensor. However, while the FA is correct, the V1 directions differ from tensor2metric, so more work is required to perfectly emulate these tools.

![mrtrix](https://github.com/niivue/niivue/assets/8930807/0dd97771-cf1b-4817-9e78-cc134ca26aa5)

neurolabusc:
The current version will not derive V1 and FA metrics from the MIF [symmetric matrices](https://community.mrtrix.org/t/dti-volumes-storage-formats-and-conversion/4502) created by [dwi2tensor](https://mrtrix.readthedocs.io/en/dev/reference/commands/dwi2tensor.html). Rather, it supports V1 and FA MIF images created using [tensor2metric](https://mrtrix.readthedocs.io/en/dev/reference/commands/tensor2metric.html). The compact and fast methods are unstable, so robust support requires either slow methods or complex methods. C solutions are provided and described [here](https://www.mpi-hd.mpg.de/personalhomes/globes/3x3/). 

The Python code below provides a solution for a single voxel created by dwi2tensor (included in `tensors.mif`). The Python code provides the correct solution, and provides a challenging test case for JavaScript solutions.

We can revisit this at a future date, but we need to decide how much code or dependencies should be provided to support this edge case.


```python
import numpy as np

def calculate_fa_and_principal_vectors(tensor_coefficients):
    # Unpack the tensor coefficients
    Dxx, Dyy, Dzz, Dxy, Dxz, Dyz = tensor_coefficients
    
[...]

PR #1006, opened 2024-06-24
Improve MIF support

neurolabusc:

- #1005


src/nvimage/index.ts
```diff
@@ -1902,6 +1977,51 @@ export class NVImage {
         } // for z
       } // for t (time)
     } // for d (direction, phase/real, etc)
+    /*
```
hanayik, line 1980: @neurolabusc , do you want to keep all of these commented lines in the code?

neurolabusc: @hanayik I do think we should keep the commented lines for a bit. The idea here is that we could mimic the features of [tensor2metric](https://mrtrix.readthedocs.io/en/dev/reference/commands/tensor2metric.html) to derive FA, ADC, V1, etc. from the images created by [dwi2tensor](https://mrtrix.readthedocs.io/en/dev/reference/commands/dwi2tensor.html). The challenge is to find a compact, fast and stable function for this. I describe this in [issue 1005](https://github.com/niivue/niivue/issues/1005). My sense is we keep the commented code for a bit, providing all the boilerplate to support different tensor2metric functions.

hanayik (approved):

PR #1029, opened 2024-07-11
implement multiplanarShowRender

### LINKED ISSUE #1024: allow developers to hide 3D render in multiplanar layout mode
Currently, NiiVue detects the aspect ratio of the canvas and chooses to show the 3D render tile if space allows. Even if space doesn't allow, the option `opts.multiplanarForceRender` can be used to force the 3D render tile to be visible. 

However, there is no equivalent option to force the 3D render tile to _never_ show the 3D render in multiplanar layout mode. 

I propose a new option: `opts.multiplanarShowRender`. By default the value can be `true` to preserve the current behaviour, but users can set this to `false` to ensure the 3D render tile never gets drawn even if the canvas aspect ratio allows for it. 

I imagine this should be implemented in [`drawSceneCore`](https://github.com/niivue/niivue/blob/04802ee7f32ed28290da32250aa7939f8a1717cc/src/niivue/index.ts#L10019)

hanayik:
This PR implements the new option `multiplanarShowRender`. 

The main motivation for this new option is that the existing option, `multiplanarForceRender`, did not allow full control over the render tile slot when niivue is in multiplanar layout mode. 

I have implemented this in a way that preserves backwards compatibility with `multiplanarForceRender`. 

The logic is:
1. check if the user set `multiplanarForceRender=true`. 
2. if `multiplanarForceRender === true` then assume the user/developer has no knowledge of the new option `multiplanarShowRender`. NiiVue behaviour is identical to before. 
3. if `multiplanarForceRender === false` (the default value) then we inspect the value of `multiplanarShowRender`. 
4. if `multiplanarShowRender === 'always'` then we draw the render tile in multiplanar mode
5. if `multiplanarShowRender === 'never'` then we leave the render tile blank
6. if `multiplanarShowRender === 'auto'` (the default value) then we only draw the render tile if the NiiVue canvas aspect ratio allows us that space for free (this is how NiiVue worked prior to this change as well). 

So, basically, if someone sets `multiplanarForceRender=true` then we operate as if the new option does not exist. There is no behaviour change in NiiVue. 

If `multiplanarForceRender=false` and `multiplanarShowRender` is anything other than `auto` then we obey the user's wishes and do whatever `multiplanarShowRender` suggests.  

This backwards compatible implementation does not require any demo html updates since those will still work as before. 

- closes #1024

### DIFF
--- a/src/niivue/index.ts
+++ b/src/niivue/index.ts
@@ -10160,6 +10160,27 @@ export class Niivue {
         this.draw2D([0, 0, 0, 0], 2)
       } else {
         // sliceTypeMultiplanar
+        let showRender = false
+        // this.opts.multiplanarForceRender is deprecated but was boolean.
+        // We now need to check if it is true. If so, then the user may not know
+        // about the new multiplanarShowRender option, so we need to show the render.
+        if (this.opts.multiplanarForceRender) {
+          showRender = true
+          // warn the user that the option is deprecated
+          log.warn(
+            'multiplanarForceRender is deprecated. Use multiplanarShowRender instead. Possible values are: "always", "auto", "never".'
+          )
+        } else {
+          //  check the now preferred multiplanarShowRender option.
+          // if the value is always, then show the render.
+          if (this.opts.multiplanarShowRender === 'always') {
+            showRender = true
+            // again, warn the user that we are using the new option
+            log.warn(
+              'multiplanarShowRender is set to always and multiplanarForceRender (deprecated) is false. We are assuming you prefer the non-deprecated option: multiplanarShowRender.'
+            )
+          }
+        }
         const isDrawPenDown = isFinite(this.drawPenLocation[0]) && this.opts.drawingEnabled
         const { volScale } = this.sliceScale()
         if (typeof this.opts.multiplanarPadPixels !== 'number') {
@@ -10208,7 +10229,7 @@ export class Niivue {
         }
         if (isDrawColumn) {
           let ltwh = ltwh1x3
-          if (this.opts.multiplanarForceRender || ltwh1x4[4] >= ltwh1x3[4]) {
+          if (showRender || (this.opts.multiplanarShowRender === 'auto' && ltwh1x4[4] >= ltwh1x3[4])) {
             ltwh = ltwh1x4
           } else {
             isDraw3D = false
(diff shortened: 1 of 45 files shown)


neurolabusc:
@hanayik in general this looks fine. However, I would change the behavior for when the old binary `multiplanarForceRender` is used, so that it converts this to the new ternary value and discards the obsolete option. This will purge the old property if the user saves the scene as a document. Indeed, while I leave it in the sample before, I would even remove the log.warn() as the new property is a superset of the old property, so old scenes will just work. 

```
if (this.opts.multiplanarForceRender) {
    log.warn('multiplanarForceRender is deprecated. Use multiplanarShowRender instead. Possible values are: "always", "auto", "never".')
    if (this.opts.multiplanarForceRender) {
        this.opts.multiplanarShowRender = 'always'
    } else {
        this.opts.multiplanarShowRender = 'auto'
    }
    delete this.opts.multiplanarForceRender;
}
```

src/niivue/index.ts
```diff
@@ -10160,6 +10160,27 @@ export class Niivue {
...
+        if (this.opts.multiplanarForceRender) {
+          showRender = true
+          // warn the user that the option is deprecated
+          log.warn(
+            'multiplanarForceRender is deprecated. Use multiplanarShowRender instead. Possible values are: "always", "auto", "never".'
+          )
+        } else {
+          //  check the now preferred multiplanarShowRender option.
+          // if the value is always, then show the render.
+          if (this.opts.multiplanarShowRender === 'always') {
```
cdrake, line 10176: We should define an enum for this.

src/niivue/index.ts
```diff
@@ -10160,6 +10160,27 @@ export class Niivue {
...
+            showRender = true
```
cdrake, line 10177: This should be renamed with "is", like isShowingRender to clarify that this is a boolean.

cdrake (changes requested):

hanayik:
@neurolabusc and @cdrake, I have implemented your suggestions and tested. Once the tests pass in github I will merge this and release a new version so the feature is published. I am eager to use this new option in one of my projects.

neurolabusc:
@hanayik I committed a PR directly into your refactor branch that updates the live demos to use the modern usage. It also addresses a regression with legends that are wider than the canvas.

I updated the live demo recipes to remove deprecated property. The end to end tests still include the deprecated property. Most of these should probably be updated as well (though perhaps you want to keep one to ensure that we still support deprecated usage).

With these changes, I approve your PR and I encourage you to merge this PR if my changes meet your standards.

PR #1033, opened 2024-07-15
allow concurrent fetch in loadVolumes and loadMeshes

### LINKED ISSUE #1032: enhancement: allow concurrent fetch when loading volumes and meshes by URL
The NiiVue methods `loadVolumes` and `loadMeshes` use the `fetch` API to load images from URLs. However, those methods take a list of images and fetch them serially. 

These methods can be enhanced to fetch multiple images concurrently and then once fetched add them all to the scene.

hanayik:
This PR updates `loadVolumes` and `loadMeshes` to concurrently fetch image data using the `fetch` API. This results in much faster load times for scenes where multiple images are fetched from a remote server. 

In the minimal tests below, the concurrent `fetch` implementation results in roughly a **1.8x speedup** to fetch all image data from an S3 bucket. 

## Previous behaviour

Prior to this PR, NiiVue would load images serially. You can verify this by monitoring the network tab in the developer window of the Chrome web browser. 

You can see that the browser doesn't even finish fetching image 4 until just **after 2500 ms**.

Screenshot 2024-07-15 at 15 58 57

## New behaviour

Opening the network tab in the developer window with the new implementation shows that all of the `fetch` requests start at the same time. This leads to faster initial scene renders since we do not have to wait on serial `fetch` requests. 

The concurrent `fetch` calls all finish **within 1400 ms**

Screenshot 2024-07-15 at 15 58 17

- closes #1032 
- note: previously, NiiVue would render each image as soon as it was fetched and parsed. This meant that you could visually see images load and "appear" in the canvas (more obvious with large images or slow networks). The new implementation will not render any images until all are fetched and parsed. There is no functional downside to this, but the behaviour may "appear" different when watching a scene load over a network connection.


hanayik:
I have tested this PR locally on my Ubuntu machine and all tests pass. I think the github actions Linux environment is to blame for the failed tests. I'm merging this so that the feature can be released.

PR #1051, opened 2024-08-21
Add user defined clickable labels

cdrake:

- [Allow users to add anchored labels with user defined click handlers](https://github.com/niivue/niivue/issues/1050)
-
-


taylorhanayikVBC (changes requested):
@cdrake , `this.drawAnchoredLabels()` is never called anywhere, so the anchored labels are never appearing even though they are setup in the new `labels.anchored.html` demo. Perhaps this should be added to `drawSceneCore` or somewhere similar?

cdrake:
@taylorhanayikVBC  - Yes, that was main issue.  I added the ```this.drawAnchoredLabels()``` call to ```drawSceneCore``` and added a test to verify correct rendering.

PR #1069, opened 2024-09-16
Pluri/features

hanayik:
This (rather large) PR implements multiple new features and enhancements to existing niivue features.

Most of these changes are backwards compatible with the previous niivue version, but a subset of features are not (mainly due to function argument changes).

## clickToSegment (aka magic wand tool)
- can now automatically determine whether to segment bright or dark clusters
- accepts user supplied intensity thresholds for stopping segmentation 
- can now be limited by distance in millimeters to avoid including voxels far away
- some flood-fill enhancements to enable these features too
- allows changing pen color to use clickToSegment multiple times with different pens

## drawing pen
- now can use `penSize` value for thicker strokes when drawing freehand 
- add ability to do binary cluster dilation (grow clusters from drawings or from clickToSegment)

## broadcastTo (syncing)
- allow choosing specific properties to broadcast to other niivue instances
- reimplement logic to make it more maintainable and extensible

## ROI selection tool
- add new shader for ROI selection dragMode
- implement new dragMode `roiSelection`

## descriptive stats
- updated `getDescriptives` to use ROI region as input mask
- compute area when ROI is a 2D slice
- change `getDescriptives` input argument to an object rather than many positional arguments


neurolabusc:
@hanayik the code looks fine, I think that you should have a live demo in /demos/features/ and remove the file src/index_segment.html - this approach would allow our documentation to link to a hosted web page.

PR #1083, opened 2024-10-20
remove unstyled-button class from features in demo project

adambasha0:
unstyled-button in afni, timeseries, and timeseries2 files in features folder were missing semicolon. After the fix, I found out the class was redundant and not needed any more. Therefore it was removed

### DIFF
--- a/demos/features/afni.html
+++ b/demos/features/afni.html
@@ -6,18 +6,10 @@
     <meta name="viewport" content="width=device-width,initial-scale=1.0" />
     <title>AFNI HEAD/BRIK format</title>
     <link rel="stylesheet" href="niivue.css" />
-    <style>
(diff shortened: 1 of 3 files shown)


adambasha0:
after_and_before_fix_timeseries2

adambasha0:
fix_behaviour

adambasha0:
prefix_behaviour

hanayik:
@adambasha0 , could you update your PR to this implementation instead? Modify as needed to suit the separate demos.

```
<label for="check1">Normalize graph</label>
<input
  type="checkbox"
  id="check1"
  value="true"
/>
```

hanayik:
closing this since we will be phasing out these demos as we transition to docusaurus.

PR #1097, opened 2024-11-07
set parent div background color to opts.backColor

hanayik:
This PR changes how we set the canvas parent background color. We set it to black by default previously, but I think it makes more sense to set the canvas's parent background color to the same color specified in `opts.backColor`. This way, a div with padding around the canvas wont show a black border behind the canvas.


hanayik:
Closing for now. We can revisit later. This can always be changed by downstream developers after attaching niivue to the canvas

Issue #823, opened 2024-01-12
use parameter objects when more than 3 arguments to a function


hanayik:
originally from #769:

### Methods with many parameters: use parameter objects
Some methods inside NiiVue have tons of parameters. This means that users have to pay attention to the ordering of the parameters and makes optional parameters harder, as they have to be at the end of the parameter list. Instead, I would recommend using parameter objects (like loadFromFile in NVMesh) for methods with more than 3 parameters.

TejasNangru:
@hanayik Hi, You told me to choose five functions that could be updated to use a parameter object within this issue,
Can i choose any random five functions from any file or is there any particular files on which i have to focus on?(Please give me their names(if particular))

TejasNangru:
@hanayik sir, any updates for it, please?

hanayik:
@TejasNangru , you can highlight all offending functions that have more than 3 input arguments by adding a new rule to the file `niivue/packages/niivue/.eslintrc.cjs`:

Just copy and paste this to replace the "overrides" section of that file

```
overrides: [
    {
      files: ['src/**/*.ts'],
      parserOptions: {
        project: ['./tsconfig.json']
      },
      rules: {
        '@typescript-eslint/explicit-function-return-type': 'error',
        '@typescript-eslint/no-floating-promises': 'error',
        '@typescript-eslint/no-misused-promises': 'error',
        // allow any for now: TH implemented for CR easier development cycles
        '@typescript-eslint/no-explicit-any': 'off',
        'max-params': 'off',
        '@typescript-eslint/max-params': 'error'
      }
    }
  ]
```

if you use a modern IDE or code editor, it should show you errors at all function locations where there are more than 3 arguments.

TejasNangru:
ohk, sorry for late reply, I was busy in my college exams,
Now I will continue working on this.

neurolabusc:
@TejasNangru we appreciate any contributions from the community to improve NiiVue. However, if you are specifically interested in becoming a GSoC 2025 contributor, do please [see our advice for GSoC 2025](https://github.com/niivue/niivue/issues/1192#issuecomment-2708939315).

TejasNangru:
okay sir!

PR #1098, opened 2024-11-18
Converted UI Components to use config object in constructor

cdrake:

- Use parameter objects for component construction and draw calls [823](https://github.com/niivue/niivue/issues/823)
- Updated associated demos
- Removed button container demo from index.html
- Updated tests


hanayik (approved):
@cdrake , this looks really clean. I like the config interfaces. We should do this for all of Niivue (as a future PR) following this work as a template.

PR #1100, opened 2024-11-19
Fixed missing legend text

cdrake:

- [Fixed missing legend text](https://github.com/niivue/niivue/issues/1099)


hanayik (approved):
@cdrake , this looks good to me. However, I was reviewing on my iPad using github codespaces to run the demos, and noticed that touch events are not being handled correctly anymore. I’ll merge this PR, but we’ll need a separate fix and new PR to solve the touchEvent handling. 

The error with touch events is that they are not only being observed by the canvas, but they are also causing scrolling on the page. Maybe it is a side effect of the new `convertTouchEventToPointerEvent` function that was introduced.

Issue #1139, opened 2025-01-13
smaller values are not highlighted


nicoalee:
Hi Niivue team! While using Niivue to display ROIs in a brain map, I came across a potential bug.

Using the alphathreshold demo, you can see that setting the +threshold to 0 yields a different visual result depending on the upper bound set (i.e. cal_max). This appears to be a bug as the expected behavior would be that both images should yield the same activations when the threshold is 0 (albeit with differences in coloration).

 
<img width="1552" alt="Screenshot 2025-01-13 at 12 32 32" src="https://github.com/user-attachments/assets/8e226871-16a4-45a7-a2c4-de353f928ce9" />

<img width="1552" alt="Screenshot 2025-01-13 at 12 35 12" src="https://github.com/user-attachments/assets/5b824640-1fb5-4d70-94c9-7e8e18a0c56e" />

neurolabusc:
Yes, this was an oversight in the demo. Statistical values should not touch or cross zero for the outline shader. The zero point is the no-mans land that is neither correlated nor anti-correlated. The actual outline is comuted in the graphics shader. I have changed the live demo to have the slider go from t >= 0.1..5.0 instead of 0.0..5.0. The latest code also logs a warning when the outline shader is used with thresholds that touch or cross zero.

nicoalee:
The same kind of behavior can be seen even if `t >= 0.1` (so maybe this is a rounding issue?)

For example: Here is a fork of the alphathresh demo with cal_min set to 0.1, and cal_max set to 6.
<img width="1552" alt="Screenshot 2025-01-13 at 19 55 35" src="https://github.com/user-attachments/assets/8b02fa08-2df8-48c6-9b4b-5b79996ce132" />

Here is the same example where cal_min is also set to 0.1, and cal_max is set to 60
<img width="1552" alt="Screenshot 2025-01-13 at 19 55 29" src="https://github.com/user-attachments/assets/02f3e114-6eb5-4a65-82ee-979dcc2023ef" />

Intuitively, one would expect that both these images would have the same outlined activation areas, with differing color map intensity. However we can see that fewer areas are outlined and shown even though the cal_min threshold value has not changed.

neurolabusc:
[...]

PR #1145, opened 2025-01-15
Ensure that voxel that survives threshold is not rounded to transpare…

neurolabusc:
Voxels that survive threshold should be rounded up to survive threshold. Once merged, the [alphathreshold live demo](https://niivue.github.io/niivue/features/alphathreshold.html) will expose more features

- #1139

### DIFF
--- a/packages/niivue/demos/features/alphathreshold.html
+++ b/packages/niivue/demos/features/alphathreshold.html
@@ -12,14 +12,14 @@
       <strong>niivue requires JavaScript.</strong>
     </noscript>
     <header>
+      <label for="dpiCheck" title="higher resolution for 'retina' displays">HighDPI</label>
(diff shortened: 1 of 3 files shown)


packages/niivue/demos/features/alphathreshold.html
```diff
@@ -138,13 +155,17 @@
       nv1.setSliceMM(false)
       nv1.opts.isColorbar = true
       await nv1.loadVolumes(volumeList1)
+      slideOutline.oninput()
       nv1.volumes[0].colorbarVisible = false //hide colorbar for anatomical scan
-      nv1.volumes[1].alphaThreshold = true
+      //nv1.volumes[1].alphaThreshold = false
```
cdrake, line 160: We should remove this if it is unused

cdrake (approved):
LGTM

PR #1151, opened 2025-01-21
Create color picker control. Consolidate general and options tabs

cdrake:

- Extracts out color picking into a separate control
- Consolidate General and Options tabs


neurolabusc:
Looks good

PR #1153, opened 2025-01-22
implement system provided html color picker

hanayik:
This PR implements the ColorPicker using standard HTML input elements. This allows the System/browser to display its native color picker supplied by the framework or OS.

Importantly, this does not have overflow side effects in the desktop app UI. The color picker is displayed over the browser window and can even extend beyond the electron browser window. 

The only missing feature is Alpha channel support. This can be overcome by adding a slider specifically for alpha if we want it. This would be added to the ColorPicker component.


cdrake (approved):
LGTM

PR #1155, opened 2025-01-23
Changed visibility icons to radix-ui react icons added visibility icons to mesh layer card.

cdrake:

- Changed visibility icons to radix-ui react icons, removed @mui dependencies


neurolabusc (approved):
Looks good

PR #1164, opened 2025-01-28
Add enum change control (Slice Type and Multiplanar Layout)

cdrake:

- Adds enum change controls for Slice Type and Multiplanar Layout
- Adds gamma getter/setter for document


hanayik:
@cdrake , can you merge #1166 when all checks pass, and then rebase your branch on main? The menu bar items should work now, and you can remove the sliceType and layout UI components you added to the desktop in favour of using the menu bar actions. There were already menu bar entries for those (I just didn't know they were not visible on other platforms). Sorry about that!

PR #1166, opened 2025-01-28
enable menubar on all platforms; lint

hanayik:
This PR ensures that the application menu bar appears on all platforms. I have also fixed code formatting in the desktop package by running `npm run lint`. 

@cdrake , this should fix the menu bar issue you were having. I would recommend that we merge this into main, and then you rebase your branch on main.


cdrake (approved):
LGTM

PR #1172, opened 2025-01-30
Order volumes

cdrake:

- Allow users to reorder volumes
- Add EnumSelect to update enums in NVConfigOptions


neurolabusc (approved):
Looks good

PR #1175, opened 2025-02-04
Fixes duplicate volumes or meshes being added

cdrake:

- Fixes duplicate volumes and meshes being added on state updates


neurolabusc (approved):
Great

PR #1178, opened 2025-02-06
A higher order gradient estimation option and gradient rendering

thewtex:
@neurolabusc looking to visually compare the impact of different gradient methods (first row) and the gradients themselves (second row). Do you have any ideas or suggestions on how to visualize the gradients. E.g.

- Use the `gradientTexture` as the `volumeTexture`?
- Extract the `gradientTexture` volume, then apply as an overlay volume?
- [...]


neurolabusc:
See my [PR](https://github.com/thewtex/niivue/pull/1) for your WIP.

thewtex:
Should `renderGradientValues` be a member of `NVConfigOptions`?

neurolabusc:
@cdrake is the ultimate maintainer for new configurations, as we want to make sure they get saved in the scene. 

My own sense is that creating a new option `renderGradientValues ` will confuse users, as one needs to establish precedence of this option over the `gradientTextureAmount` (where 0 means volume rendering without matcap-based lighting illumination, and values greater than 0 but less than 1 set the mix better the pure volume rendering and the matcap texture. My sense is to use the existing `gradientTextureAmount` to choose between the three shaders, e.g. if `NaN` (or infinity, or some other value) you render the gradients directly, if 0 you use the volume renderer, and if >0 and <= 1 you use the matcap volume rendering. Providing a single option with multiple settings seems cleaner than multiple options where one overrides the other.

thewtex:
It seems:

- We may need 16-bit precision in `TEXTURE8_GRADIENT_TEMP` for the second-order case
- For the first-order case, we can optimize `TEXTURE8_GRADIENT_TEMP` with `r8Tex` instead of `rgbaTex`?

neurolabusc:
@thewtex  
r8Tex() creates a R8 texture (8bits, one component image). We currently create TEXTURE8_GRADIENT_TEMP using rgbaTex() creates a RGBA8 texture (32-bits, 8-each for R,G,B,A). We use all for to map the gradient (RGB maps to XYZ direction, A maps to gradient amplitude).

If you need more precision, you could have TEXTURE8_GRADIENT_TEMP be a [RGBA32UI](https://webgl2fundamentals.org/webgl/lessons/webgl2-whats-new.html) with no other changes. Using RGBA32I would be more intuitive, but you will need to make some changes in the shaders, e.g. removing the line:

```
grad.rgb = normalize(grad.rgb*2.0 - 1.0);
```

hanayik (approved):

PR #1186, opened 2025-02-18
implement new drag mode for adjusting window and level

hanayik:
This PR introduces new functionality that allows users to set the "drag mode" of the primary mouse button (the left button) and for single finger touch drags. 

we have a new `DRAG_MODE_PRIMARY` enum with the options:
- `DRAG_MODE_PRIMARY.windowing`
- `DRAG_MODE_PRIMARY.crosshair` (this is the default to preserve existing functionality)

What does the new windowing click and drag movement do?
- click and drag up in the canvas to make things brighter (moves the display range uniformly lower)
- click and drag down to make things darker (moves the display range uniformly higher)
- click and drag left to increase the contrast (makes the display range more narrow)
- click and drag right to decrease the contrast (makes the display range wider)
- These movements also work in volume rendering (3D) mode, allowing for interesting effects

users can set the primary drag mode in the niivue constructor by passing in the option object like this:

```javascript
import { Niivue, DRAG_MODE_PRIMARY} from '@niivue/niivue'

const nv = new Niivue({
  dragModePrimary: DRAG_MODE_PRIMARY.windowing // or crosshair
})
```

If you want to change the drag mode any time after initializing niivue, then you can simply set `nv.opts.dragModePrimary = DRAG_MODE_PRIMARY.crosshair // or .windowing`

If the windowing drag mode is active, the user can override it by holding the `control` key during the click and drag action with the mouse. This allows using the crosshair mode temporarily. 

Lastly, DRAG_MODE in NiiVue refers to the secondary mouse button (right button). A new enum has been made available in NiiVue called DRAG_MODE_SECONDARY that replicates the original DRAG_MODE enum. It would be nice to deprecate DRAG_MODE eventually in favour of the new DRAG_MODE_SECONDARY and DRAG_MODE_PRIMARY enums.

### DIFF
--- a/packages/niivue/src/niivue/index.ts
+++ b/packages/niivue/src/niivue/index.ts
@@ -1647,7 +1647,6 @@ export class Niivue {
     const valmin = this.volumes[volIdx].cal_min
     const valmax = this.volumes[volIdx].cal_max
     const level = (valmax - valmin) / 2 + valmin
-    console.log(valmin, level, valmax)
     this.refreshLayers(this.volumes[volIdx], 0)
     // set the current mouse position (window space) as the new reference point
     // for the next comparison
--- a/packages/niivue/src/niivue/index.ts
+++ b/packages/niivue/src/niivue/index.ts
@@ -73,7 +73,8 @@ import {
   Scene,
   SLICE_TYPE,
   SHOW_RENDER,
-  DRAG_MODE,
+  DRAG_MODE, // DRAG_MODE_SECONDARY is the same as DRAG_MODE. DRAG_MODE may be deprecated.
+  DRAG_MODE_PRIMARY,
   COLORMAP_TYPE,
   MULTIPLANAR_TYPE,
   DEFAULT_OPTIONS,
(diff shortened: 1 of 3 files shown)


packages/niivue/src/niivue/index.ts
```diff
@@ -1581,6 +1593,68 @@ export class Niivue {
...
+      // ensure min is not above max
+      mn = mx - 1
+    }
+
+    this.volumes[volIdx].cal_min = mn
+    this.volumes[volIdx].cal_max = mx
+    const valmin = this.volumes[volIdx].cal_min
+    const valmax = this.volumes[volIdx].cal_max
+    const level = (valmax - valmin) / 2 + valmin
+    console.log(valmin, level, valmax)
```
neurolabusc, line 1650: Either remove console.log() or use logger.

neurolabusc (changes requested):
Primary code looks good. Please remove console.log() or use logger

neurolabusc (approved):

PR #1200, opened 2025-03-05
experimental zarr support

hanayik:
This draft PR implements a VERY simple zarr loader for niivue using [zarrita](https://zarrita.dev/packages/zarrita.html) from @manzt. The first pass PoC basically copies the zarrita observable notebook. 

@neurolabusc , and @thewtex: This is a work in progress. It is intended to be a starting point for adding zarr loading support in niivue. All contributions and feedback are welcome. It will definitely take a concerted effort to get this over the finish line. 

to run this locally:

- clone the niivue repo
- `git checkout hanayik/experimental-zarr`
- `npm install`
- `npm run dev`

todo:
- Clean up messy code
- refactor `NVImage.new` to use a parameter object
- support loading more than just `kind: array`
- query number of chunks and selectively load certain chunks
- only load chunks covering the current viewport region
- only load chunks at the current zoom level


packages/niivue/src/nvimage/index.ts
```diff
@@ -3352,6 +3378,42 @@ export class NVImage {
...
       }
     }
+    const re = /(?:\.([^.]+))?$/
+    let ext = ''
+    // if zarr
+    ext = re.exec(url)[1]
+    // try url and name attributes to test for .zarr
+    if (ext === 'zarr' || re.exec(name)[1] === 'zarr') {
+      const store = new zarr.FetchStore(url)
+      const arr = await zarr.open(store, { kind: 'array' })
```
manzt, line 3388: NB: zarrita has some [type-level helpers](https://zarrita.dev/packages/zarrita.html#data-typing-in-typescript) that I sprinkle in often with some assertions to make some high-level checks about the data source. For instance, in the code below, object, string, or bigint arrays would throw an error.

```ts
arr // zarr.Array<zarr.DataType, zarr.FetchStore)>
if (!arr.is("number")) {
  throw new Error(`Unsupported zarr data_type, got: ${arr.dtype}.`)
}
arr // zarr.Array<"int8" | "int16" | "int32" | "uint8" | "uint16" | "uint32" | "float32"| "float64", zarr.FetchStore>
```

The `.is` method hooks into TypeScript's type narrowing, and the resulting `chunk.data` from `zarr.get` will be the appropriate union type of TypedArrays.

PR #1206, opened 2025-03-06
fix path

hanayik:
@thewtex , this PR fixes the path to the bundled niivue index.js in the new gradient demo page.


thewtex:
@hanayik thank you!!

PR #1222, opened 2025-03-13
Center the thumbnail on the canvas

thewtex:
Removes large visual shifts in content with the thumbnail and
interactive rendering.


thewtex:
How the test case renders before (left) and after (right):

image

hanayik (approved):

PR #1227, opened 2025-03-18
await attachToCanvas calls

thewtex:
Ensure initialization completes before subsequent calls.

### DIFF
--- a/docs/development-notes/dicom-manifests.md
+++ b/docs/development-notes/dicom-manifests.md
@@ -19,7 +19,7 @@ var volumeList1 = [
 ];
 var nv1 = new Niivue();
 nv1.setRadiologicalConvention(false);
-nv1.attachTo("gl1");
+await nv1.attachTo("gl1");
 nv1.loadVolumes(volumeList1);
 nv1.setSliceType(nv1.sliceTypeMultiplanar);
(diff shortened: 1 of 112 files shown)


neurolabusc:
@thewtex nice catch, I had never noted that was async. I do think you will also want to add and await [here](https://github.com/niivue/niivue/blob/c48bd8809a29b47d0cf62091f84de6d1b2cf028e/packages/niivue/src/niivue/index.ts#L4135):

```javascript
await nv1.attachTo("${canvasId}");
```

hanayik:
@thewtex , it pains me to suggest this, but I think it would be good to make this change as comprehensive as possible. 

Running this command shows 100 additional occurrences that need an "await".

```
# from the root of the niivue repo
cd packages/niivue

# check for un-awaited attachTo calls
grep -r "attachTo.*" \
  --include="*.ts" \
  --include="*.html" \
  --exclude-dir="dist" \
  --exclude-dir="build" \
  src playwright demos \
| grep -v "await nv.*\.attachTo"
```

Do you mind doing a bulk fix in your fork?

thewtex:
@hanayik :+1: done.

hanayik (approved):

PR #1251, opened 2025-04-03
Start cleanup method

thewtex:
This will remove the window 'resize' event observer and disconnect the
resizeObserver.

A MutationObserver is also added for the attached canvas's parent so
this method is called whenever the canvas associated with a Niivue
instance is removed.

This cleans up references / addresses memory leaks, and e.g. prevents invalid
calls to this.resizeListener with a window resize after the associated
canvas has been removed from the DOM.

### DIFF
--- a/packages/niivue/src/niivue/index.ts
+++ b/packages/niivue/src/niivue/index.ts
@@ -431,6 +431,8 @@ export class Niivue {
   extentsMax?: vec3
   // ResizeObserver
   private resizeObserver: ResizeObserver | null = null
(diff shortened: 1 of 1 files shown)


hanayik (changes requested):
@thewtex , this looks great. I think it would make sense to unregister the [event listeners](https://github.com/niivue/niivue/blob/7af93b4fd85ed7a0f9c888aeddff7d5ce510b8f5/packages/niivue/src/niivue/index.ts#L2023) that are setup in `registerInteractions` as well.

hanayik (approved):

PR #1267, opened 2025-04-10
Add custom layout functionality and related methods to Niivue

### LINKED ISSUE #1237: [Feature Request] Support Custom Viewer Layouts
I'd like to request a feature that allows customizing the layout of viewers.

In our project, we need to display three viewers in a specific layout like the one below. As far as I know, NiiVue doesn't currently support a customizable layout feature.

Image

So, what we're doing now is creating three separate NiiVue instances, one for each viewer, and setting different SLICE_TYPE options to achieve the layout. But I feel like this is more of a workaround than a proper solution — mainly because it’s likely not memory-efficient or performant, since I assume each instance doesn’t share volume data.

It would be great if there were an API like this, where we can pass in a canvas element and SLICE_TYPE, and it renders the appropriate slice on that canvas:
```
niivue.attachTo(canvasElement1, SLICE_TYPE.AXIAL)
niivue.attachTo(canvasElement2, SLICE_TYPE.SAGITTAL)
niivue.attachTo(canvasElement3, SLICE_TYPE.CORONAL)
```

hanayik:
- Introduced `setCustomLayout`, `clearCustomLayout`, and `getCustomLayout` methods.
- Implemented custom layout handling in `drawSceneCore`. Custom layouts have higher priority than built-in layouts.
- Updated hot-reloadable index.html to include a toggle button for custom layouts during development.

### Example:

```javascript
// Layout format: {sliceType: SLICE_TYPE, position: [left, top, width, height]}
// Units are relative (0-1) to canvas size
const customLayout = [
  // Left 50% - Sag
  {sliceType: SLICE_TYPE.SAGITTAL, position: [0, 0, 0.5, 1.0]},
  // Top right - Cor
  {sliceType: SLICE_TYPE.CORONAL, position: [0.5, 0, 0.5, 0.5]},
  // Bottom right - Ax
  {sliceType: SLICE_TYPE.AXIAL, position: [0.5, 0.5, 0.5, 0.5]}
];
      
  // Apply the custom layout
  nv1.setCustomLayout(customLayout);

  // clear the custom layout and revert to built-in layouts
  nv1.clearCustomLayout()

/*
   * +----------------+----------------+
   * |                |                |
   * |                |     coronal    |
   * |                |                |
   * |                |                |
   * |   sagittal     +----------------+
   * |                |                |
   * |                |     axial      |
   * |                |                |
   * |                |                |
   * +----------------+----------------+
 */
```

image

- closes #1237

### DIFF
--- a/packages/niivue/src/index.html
+++ b/packages/niivue/src/index.html
@@ -188,7 +188,7 @@
       }
       let defaults = {
         // backColor: [0, 0.2, 0.4, 1],
-        backColor: [0,0,0,1],
+        backColor: [0, 0, 0, 1],
         show3Dcrosshair: true,
         loglevel: 'debug',
         isRuler: true,
@@ -199,7 +199,8 @@
         showMeasureUnits: true,
         measureTextColor: [0, 1, 0, 1],
         measureLineColor: [1, 0, 0, 1],
-        measureTextHeight: 0.04
+        measureTextHeight: 0.04,
+        isColorbar: true
       }
       var nv1 = new Niivue(defaults)
       await nv1.attachToCanvas(gl1)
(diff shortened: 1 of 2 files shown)


neurolabusc:
This is very nice. However, you should deduct the size of the colorbar (if present) from the size of the canvas. You can do this with reserveColorbarPanel() to determine canvas pixels required for colorbar.

You can see the issue by setting the `isColorbar` to true: 

```javascript
      var nv1 = new Niivue(defaults)
      await nv1.attachToCanvas(gl1)
      nv1.opts.isColorbar = true
```

hanayik:
@neurolabusc , good catch! I have added the space reservation for the colorbar, and also added some checks to ensure the custom layout is a valid layout (e.g. no overlapping tiles, and prevent a tile from extending past the canvas bounds) by accident.

neurolabusc (approved):
Great feature

PR #1286, opened 2025-05-01
fix playwright config for linux

cdrake:
Headless tests were having an issue getting the GL context on Linux.

- verified playwright e2e tests run on Linux
- verified playwright e2e tests run on Mac


neurolabusc (commented):
@cdrake do you really want to remove window-size, I would think this makes regression test images consistently sized?

PR #1291, opened 2025-05-09
Hanayik/histo readers

hanayik:
This PR cleans up the implementation from the previous experimental zarr branch #1200.

Now, full slices from zarr arrays can be fetched using url query strings: 
`http://localhost:5173/rgb.ome.zarr/scale0/image?z=1200`

This needs to be combined with the `name` attribute when loading images in NiiVue. 

```javascript
{
  url: 'http://localhost:5173/rgb.ome.zarr/scale0/image?z=1200',
  name: 'example.zarr'
}
```
closes #1200


neurolabusc (approved):
Great to unify this with main. Thanks!

PR #1292, opened 2025-05-11
FIX: getValue now correctly applies transform

### LINKED ISSUE #1290: Voxel intensity values are incorrectly returned
A bug was introduced in version 0.53.0, where the voxel intensity values in the location are returned in a mirror-like way, for example: [basic multiplanar](https://niivue.github.io/niivue/features/basic.multiplanar.html)
I have pinned it down to the refactor of NVImage in the file [VolumeUtils](https://github.com/niivue/niivue/commit/0d12c03bd67293347d6ff1d3889c363d4de6cc13) with the decision to assume coordinates to be in the image's native space:

```
 // Input coordinates (x, y, z) are assumed to be in the image's NATIVE space based on original function context.
  // NO reorientation logic needed here if inputs are native.
  // If coordinates were meant to be RAS, reorientation would be needed *before* this function or added here.
  // Sticking to the original implementation's apparent assumption of native coordinates.

  // Clamp coordinates to valid native range
  x = Math.max(0, Math.min(Math.round(x), nx - 1))
  y = Math.max(0, Math.min(Math.round(y), ny - 1))
  z = Math.max(0, Math.min(Math.round(z), nz - 1))
```

(You can see the value returned is 0)

Image

(Here you can see that it is returning 9, when its actually not on hovering over any voxel)

Image

I have managed to make it work properly by replacing these three variable assignments with the previous function, but decided to open this issue rather than a pull request as the refactored code seems to be the intended functionality to not convert the coordinates to RAS:

```
  // Previous function
 const perm = nvImage.permRAS!.slice()
 if (perm[0] !== 1 || perm[1] !== 2 || perm[2] !== 3) {
   const pos = vec4.fromValues(x, y, z, 1)
   vec4.transformMat4(pos, pos, nvImage.toRASvox!)
   x = pos[0]
   y = pos[1]
   z = pos[2]
}

```

(You can see the value returned is 9, which is the correct one)

Image

In both cases, with both the new and old approaches, these changes don't seem to have an impact on the [drawing example](https://niivue.github.io/niivue/features/draw2.html) as returned voxel intensity values are correct in both cases. Haven't been able to find anything in the documentation in regards to these changes or in the code pointing at how to tell NiiVue to handle either RAS or native coordinates.

hanayik:
This PR fixes a bug in the `getValue` function which was part of the NVImage and VolumeUtils refactor introduced by me. The rendering was not changed so tests missed it. I have added a new unit test to verify correct functionality going forward. 

Given the nature of the bug, I have patched and immediately released niivue v0.57.0 on NPM to get the changes out quickly. 

- closes #1290


hanayik:
@neurolabusc, I have fixed the failing unit test for `getValue`. The `hippo.nii.gz` file was problematic, and resulted in intermittent test failures for some reason. I simply unzipped and the gzipped again and it has resolved the issues. 

During my debugging, I also realised that our drawing texture needs to be recreated when we toggle between 2D and 3D textures based on the background image. Please check the logic I added.

neurolabusc (approved):

PR #1300, opened 2025-05-26
limit gradient opacity lut count to 192

thewtex:
The fragment shader can fail to compile on mobile browsers / mobile
GPU's where it is more common to run into the WebGL2 limit of 224
uniforms for fragment shaders.

xref: https://webglfundamentals.org/webgl/lessons/webgl-cross-platform-issues.html

### DIFF
--- a/packages/niivue/src/niivue/index.ts
+++ b/packages/niivue/src/niivue/index.ts
@@ -64,7 +64,8 @@ import {
   blurFragShader,
   sobelBlurFragShader,
   sobelFirstOrderFragShader,
-  sobelSecondOrderFragShader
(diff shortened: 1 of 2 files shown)


thewtex:
The `test-unit` `getValue` test was already failing in https://github.com/niivue/niivue/pull/1292

hanayik (approved):

hanayik:
@thewtex , I will fix the unit test in a separate PR. It’s unrelated, and fails on Linux in my testing, but passes on macOS. I suspect it’s the vitest environment, so I’ll probably migrate it to playwright.

Issue #1348, opened 2025-07-02
Broken documentation link


pwighton:
On the [development page](https://niivue.com/docs/develop) of the documentation, there is a link to "API documentation" which points to https://niivue.com/demos/devdocs/classes/Niivue.html which gives a page not found.

Issue #1347, opened 2025-07-02
Atlas loaded from document not aligned.


cdrake:
When a document is loaded the Atlas is not aligned with the base image.

neurolabusc:
This regression is an unintended consequence of this [feature](https://github.com/niivue/niivue/issues/1039) (mea culpa). Specifically, these [lines](https://github.com/niivue/niivue/blob/33a57b4c87a0988adb6f634535e08e727962224a/packages/niivue/src/nvimage/utils.ts#L460) where the hdr.vox_offset is now ignored.

```js
    // view.setFloat32(108, hdr.vox_offset, isLittleEndian)
    view.setFloat32(108, 352, isLittleEndian)
```

The best way to fix this is to change [this line](https://github.com/niivue/niivue/blob/33a57b4c87a0988adb6f634535e08e727962224a/packages/niivue/src/nvimage/ImageWriter.ts#L123) from
```js
hdrCopy.vox_offset = Math.max(352, hdrCopy.vox_offset)
```
to read:
```js
hdrCopy.vox_offset = 352
```
At first blush, this seems like we will lose information, but not that the padding between the end of the header and the beginning of the image is zero-filled: `const padding = new Uint8Array(paddingSize)`. So the current problem is that currently the header is written claiming a offset of 352 bytes, but images with an extended header actually have padding. A more involved solution would be to try to restore the original extended NIfTI header, but this is problematic for several reasons included leakage of private information and possible malicious code.

PR #1352, opened 2025-07-03
Cdrake/fix save document

cdrake:

- Sets dark mode as default theme
- Applies theme background color to control div for documentation option 
- Save extensions for NIfTI files
- Fixes  https://github.com/niivue/niivue/issues/1347
- FIxes  https://github.com/niivue/niivue/issues/1348


neurolabusc (approved):
Accepting will check handling of MGH files subsequently to ensure that it does not regress.

PR #1361, opened 2025-07-07
gradientGL fixes

### LINKED ISSUE #1359: readMIF problem
I've come across some issue with readMIF and I'm not exactly where in the code issue is occurring. The decompression seems to be happening correctly, but there's some issue with reading the headers for some reason and I'm not sure what exactly is going on yet.

When I load `https://niivue.com/demos/images/template.mif.gz` using loadFromUrl, the volume loads correctly. Here's an example HTML for testing purposes:
```html
<!DOCTYPE html>
<html lang="en">
  <body>
    <main id="container">
      <canvas id="gl1"></canvas>
    </main>
  </body>
</html>
<script type="module">
  import * as niivue from "https://niivue.com/demos/dist/index.js";

  var volumeList1 = [
    {
      url: "https://niivue.com/demos/images/template.mif.gz",
      colormap: "gray",
      opacity: 1,
      visible: true,
    },
  ];
  var nv1 = new niivue.Niivue({ backColor: [0.7, 0.7, 0.9, 1] });
  await nv1.attachTo("gl1");
  nv1.loadVolumes(volumeList1);
</script>
```

**However**, when I try to load the same file by fetching the image and loading it into `niivue.NVImage.new` as an ArrayBuffer, I get an error in `readMIF`:
```sh
Uncaught Error: unknown datatypeCode
    at _NVImage.readMIF (index.ts:2247:11)
    at async _NVImage.new (index.ts:544:9)
    at async test_image_1.html:59:18
```
Here's the problematic HTML for testing purposes:
```html
<!DOCTYPE html>
<html lang="en">
  <body>
    <main id="container">
      <canvas id="gl1"></canvas>
    </main>
  </body>
</html>
<script type="module">
  import * as niivue from "https://niivue.com/demos/dist/index.js";

  var nv1 = new niivue.Niivue({ backColor: [0.7, 0.7, 0.9, 1] });

  await nv1.attachTo("gl1");

  const response = await fetch("https://niivue.com/demos/images/template.mif.gz");
  const dataBuffer = await response.arrayBuffer();

  const volume = await niivue.NVImage.new(
    dataBuffer,
    "template.mif.gz",
    "gray",
    1.0  
  );

  nv1.addVolume(volume);
</script>
```

### LINKED ISSUE #1360: Undefined state for gradient shader
@thewtex identified an issue that was only observed with the following combination of features: volume uses gradient compute shader, volume uses atlas (indexed) image, operating system is Linux, and graphics card is NVidia. However, the root cause seems to be general, with most implementations providing predictable results from a state that is undefined in the implementation.

The issue is with the [gradientGL()](https://github.com/niivue/niivue/blob/594d5a754a1e46137bf5b6451fcade6523a632c0/packages/niivue/src/niivue/index.ts#L7162) compute shader that creates and releases a custom vbo, but on exit not bound to any state: 

```js
gradientGL(hdr: NiftiHeader): void {
    ...
    const vbo2 = gl.createBuffer()
    gl.bindBuffer(gl.ARRAY_BUFFER, vbo2)
    ...
    gl.deleteBuffer(vbo2)
  }
```

This is only an issue for code that calls gradientGL() and then makes a draw call before explicitly setting a VAO. However, this was not the case for refreshLayers().

The solution uses the retained genericVAO (which is a unity square for 2D or cube for 3D), but crucially exits with a retained null VAO. 

```js
  gradientGL(hdr: NiftiHeader): void {
    ...
    gl.bindVertexArray(this.genericVAO)
    ...
    this.gl.bindVertexArray(this.unusedVAO)
  }
```

Likewise, the code that calls the function does set the desired VAO prior to a draw call. As an aside, this issue is from 2019 and pre-dates NiiVue, existing in the fore-runner [MRIcroWeb](https://github.com/rordenlab/MRIcroWeb/commit/d0843b4fb97cc0f85bfa9cf3c61f7b8006ee8d7c).

neurolabusc:

- fixes #1360 
- fixes #1359

### DIFF
--- a/packages/niivue/demos/features/atlas.html
+++ b/packages/niivue/demos/features/atlas.html
@@ -109,7 +109,7 @@
         nv1.setInterpolation(this.checked)
       }
       renderMode.onchange = function () {
-        nv1.setVolumeRenderIllumination(this.value)
+        nv1.setVolumeRenderIllumination(parseFloat(this.value))
       }
       var nv1 = new niivue.Niivue({show3Dcrosshair: true, backColor: [0.5, 0.5, 0.5, 1], onLocationChange: handleLocationChange })
       await nv1.attachTo("gl1")
--- a/packages/niivue/demos/features/atlas.mgz.html
+++ b/packages/niivue/demos/features/atlas.mgz.html
@@ -70,7 +70,7 @@
         nv1.setAtlasOutline(parseFloat(this.value))
       }
       renderMode.onchange = function () {
-        nv1.setVolumeRenderIllumination(this.value)
+        nv1.setVolumeRenderIllumination(parseFloat(this.value))
       }
       var nv1 = new niivue.Niivue({show3Dcrosshair: true, onLocationChange: handleLocationChange })
       nv1.opts.multiplanarShowRender = niivue.SHOW_RENDER.ALWAYS
--- a/packages/niivue/demos/features/atlas.sparse.html
+++ b/packages/niivue/demos/features/atlas.sparse.html
@@ -15,7 +15,7 @@
         type="range"
         min="1"
         max="255"
-        value="48"
+        value="8"
         class="slider"
         id="alphaSlider"
       />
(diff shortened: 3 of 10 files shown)


neurolabusc:
@hanayik this does not completely resolve the issue of changing volume rendering shaders for NVidia cards on Linux. I am happy to have this merged, as it provides clear improvements, but it does not comprehensively close issue #1360.

Issue #1364, opened 2025-07-10
Does setBrushSize exist?


longdanhao:
I try to change the brush size using setBrushSize() after run setDrawingEnabled(). But it caused an error. 
I tried both version 0.59 and 0.61 from https://www.npmjs.com/package/%40niivue/niivue?activeTab=readme

if setBrushSize does not exist, anything else I can do to change brush size/shape (e.g., circle , square ...)?

Thanks for the help.

hanayik:
@longdanhao you can use opts.penSize = 5 // or any integer value. 

If you have a Niivue instance “nv”, then the call would be nv.opts.penSize = 5 // or some int

Hope that helps! 

https://niivue.com/docs/api/nvdocument/type-aliases/NVConfigOptions#pensize

hanayik:
The range of the penSize has no limit. You are free to supply any integer value that makes sense for your use case. You can see how penSize is used here: https://github.com/niivue/niivue/blob/1a5614b7b32534179e3b007e3cc8d7993c2458a2/packages/niivue/src/niivue/index.ts#L5532

@longdanhao , Could you share a reproducible code snippet that can be used with the development index.html setup we use? See niivue/packages/niivue/src/index.html. This is the dev UI that loads when we run `npm run dev`. This would help debug things more. 

Also, are you loading your existing drawings using the `loadDrawing` or `loadDrawingFromUrl` methods on the Niivue class? That should setup the drawing helpers for you automatically. 

Overwriting means in the current pen implementation means replacing the value at those voxels, regardless of their previous label value. The current drawing implementation doesn't have any logic to only overwrite a range of values and ignore others. As a workaround, you could separate your multi-label drawing into multiple drawing files, and toggle which one is active by using the `loadDrawing` method to load the drawing that is needed while editing certain labels. Just an idea, but maybe not appropriate depending on the app. 

https://niivue.com/docs/api/niivue/classes/Niivue#loaddrawing

longdanhao:
HI Hanayik,
Thanks for the help. I used nv.loadVolumes. Because I want to have the
voxels indices information for crosshair and I hope I can edit labels
directly to this layer. The following is my code, including how I uploade
the volumns and how i am testing the drawing function.

Really appriciate the help!

LOAD:
nv.loadVolumes([
{
url: baseFileUrl,
name: (baseFileName && baseFileName.endsWith('.nii.gz')) ? baseFileName :
"base.nii.gz",
isBackground: true,
},
{
[...]

PR #1366, opened 2025-07-11
Fix links and add recipe with penSize()

neurolabusc:

- Illustrate penSize #1364 
- Consistent case for recipes
- Fix broken URL


cdrake (approved):
LGTM

PR #1406, opened 2025-08-11
Faster 2D flood fill

### LINKED ISSUE #1405: Faster even–odd rule for 2D closed polygons
Description:

NiiVue supports closed drawing pens for manual annotations. Similarly, DICOM allows [CLOSED_PLANAR closed polygons](https://dicom.nema.org/medical/Dicom/2018d/output/chtml/part03/sect_C.8.8.6.html).

Currently, NiiVue uses a [First-In, First-Out queue-based flood fill](https://en.wikipedia.org/wiki/Flood_fill#Stack-based_recursive_implementation_(four-way)) to fill these regions. While correct, this method is suboptimal for our case: both the NiiVue pen and DICOM CLOSED_PLANAR contours are strictly 2D. In such cases, we can replace the queue-based approach with the [even–odd rule](https://en.wikipedia.org/wiki/Point_in_polygon#Ray_casting_algorithm).

For typical SNR-limited scans, both methods are fast. However, for histology-resolution images, the even–odd approach is dramatically faster. The upcoming commit adopts the even–odd algorithm, retaining the old queue method only as a commented reference for regression and performance tests. Since the queue method offers no advantages for this application, it will be removed in a future cleanup.

Benefits:
 - Significant performance improvement for high-resolution (histology-scale) images.
 - Simplified logic tailored to 2D closed contours.

Note that we maintain FIFO queues for other applications that require 3D flood fills.

neurolabusc:

- Closes #1405

### DIFF
--- a/packages/niivue/src/niivue/index.ts
+++ b/packages/niivue/src/niivue/index.ts
@@ -6407,29 +6407,27 @@ export class Niivue {
     }
   }
 
-/**
- * Fills exterior regions of a 2D bitmap using the even–odd rule, marking outside voxels with 2
- * while leaving interior voxels at 0 and borders at 1. Operates within specified bounds.
- * uses crossing number algorithm (aka even–odd rule)
- * @internal
- */
+  /**
+   * Fills exterior regions of a 2D bitmap using the even–odd rule, marking outside voxels with 2
+   * while leaving interior voxels at 0 and borders at 1. Operates within specified bounds.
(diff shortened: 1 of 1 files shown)


hanayik (changes requested):
@neurolabusc , could we move the new floodfill method to its own file in the drawing directory? Just to help the effort of organizing the code in a better way for selectively importing and using the helper functions in other packages.

PR #1454, opened 2025-11-14
Add "meshShaderIndex" support to loadMeshes()

pwighton:

- Adds support for setting `"meshShaderIndex"` when loading meshes via `loadMeshes()`, like embedded mesh loading in `loadDocument()`.


pwighton:
Thanks @hanayik, screenshot is now included.

Issue #1493, opened 2025-12-22
Performance issue when loading large local NIfTI files (nii.gz) in React + Vite vs Niivue Docs demo


ArsaCore:
## Description

I’m experiencing **significant performance lag** when loading **large local NIfTI files (nii.gz)** using Niivue in a **React + Vite** application, even when using **the exact same code** from the Niivue documentation.

What’s confusing is that the **same large NIfTI file performs very smoothly** when loaded via **drag & drop** in the official Niivue docs demo, but not in my local app.

---

## Environment

* **@niivue/niivue**: `^0.66.0`
* **React**: `19.x`
* **Bundler**: Vite
* **Build mode tested**:

  * `npm run dev`
  * `npm run build` (production)
* **OS**: Windows (issue appears environment-related rather than OS-specific)

The Niivue docs/demo appear to use:

* **React**: `^18`
* **Framework**: Docusaurus

---

## Steps to Reproduce

1. Copy the **exact code** from the Niivue docs demo:

   * Source:
     [https://github.com/niivue/niivue/blob/main/packages/docs/src/components/DrawingDemo.jsx](https://github.com/niivue/niivue/blob/main/packages/docs/src/components/DrawingDemo.jsx)
   * Only changes made:

     * Integrated into a React + Vite project
     * Ignored some TypeScript errors (no logic changes)

2. Load a small NIfTI file:

   * [https://niivue.com/demos/images/FLAIR.nii.gz](https://niivue.com/demos/images/FLAIR.nii.gz)
     Works smoothly

3. Load a **large local NIfTI file** (~66 MB, `.nii.gz`)
    Severe lag, especially when **3D view is enabled**

4. Build the app using `npm run build` and test again
    Performance issue persists in production build

---

## Key Observation

If I **drag & drop the same large local NIfTI file (66 MB)** into the Niivue viewer on this page:

[https://niivue.com/docs/drawing](https://niivue.com/docs/drawing)

…it loads **very smoothly**, including:

* 2D slice views
* 3D rendering
* Drawing tools

The performance difference is **not even comparable** to what I see in my local React + Vite app using the same Niivue demo code.

This suggests:

* The file itself is not the issue
* Niivue can clearly handle this file efficiently
* There may be an **environment / bundler / React version–related difference**

---

## Notes / Suspected Causes

I’m not sure which of these is responsible, but they seem worth mentioning:

* **React 19 vs React 18** (docs use React 18)
* **Vite vs Docusaurus** build setup

[...]

PR #1497, opened 2025-12-25
Performance enhancements

neurolabusc:

- #1493

The goal of this PR is to improve interactive performance during mouse/touch gestures. 
 - The number of draw calls per gesture is reduced. I have tried to be careful, but it is possible there are edge cases where a required draw call is omitted. Merging this will allow other developers to detect regressions.
 - The illumination model for the gradient shader uses the faster Phong rather than MatCap illumination (reducing texture reads). The goal of this shader is to reduce surface opacity, so the influence is minor. This will change the appearance of [some live demos](https://niivue.com/demos/features/gradient.opacity.html) created by @thewtex. The new shader is ~25% faster.
 - Once committed, the [voxel speed-quality tradeoff live demo](https://niivue.com/demos/features/vox.tradeoffs.html) will include a FPS counter, the ability to load a huge volume, and the option to select the gradient shader.

### DIFF
--- a/packages/niivue/demos/features/mesh.tradeoffs.html
+++ b/packages/niivue/demos/features/mesh.tradeoffs.html
@@ -10,6 +10,9 @@
   </head>
     <body>
       <header>
+        <label for="animateCheck">Animate</label>
+        <input type="checkbox" id="animateCheck"/>
+        &nbsp;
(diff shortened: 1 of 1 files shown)


thewtex:
@neurolabusc fantastic! 🚀

Issue #1464, opened 2025-12-02
Add snapping to specific views


cdrake:
Add menu items to snap to specific views to for feature parity with Surf Ice.

<img width="183" height="201" alt="Image" src="https://github.com/user-attachments/assets/685dde94-0ef9-4858-ad1b-d0c594cbf3e8" />

SandeepChauhan00:
Hi @cdrake,

I see there hasn't been activity on this for about a month. If this is still open, I'd like to implement it.

To achieve parity with Surf Ice, I propose adding a "Views" submenu to the existing context menu (right-click) containing:
- Axial (Top/Bottom)
- Coronal (Front/Back)
- Sagittal (Left/Right)

I can implement the camera logic to snap `azimuth` and `elevation` to these standard orthogonal angles.

Let me know if I should proceed!

PR #1529, opened 2026-01-27
feat: add setRenderView API for anatomical snapping

AviJxn:
### Description
This PR implements the camera logic required for **Issue #1464** ("Add snapping to specific views").

It adds a public API method `setRenderView(view: string)` to the core `Niivue` class. This method accepts standard anatomical views (`Top`, `Bottom`, `Left`, `Right`, `Anterior`, `Posterior`) and automatically calculates and sets the correct `renderAzimuth` and `renderElevation`.

### Why this helps
This backend logic is the prerequisite for adding UI menu buttons. It allows any UI layer (Context Menu, Desktop App, etc.) to simply call `nv.setRenderView('Top')` without needing to manually calculate camera angles.

### Testing
I verified this locally by running the dev server and triggering the function manually via the console.

**Screenshots of specific view snapping:**
(images below for Top, Left, Anterior, and Posterior views)

**TOP**
Screenshot 2026-01-27 213458

**LEFT**
Screenshot 2026-01-27 214728

**ANTERIOR**
Screenshot 2026-01-27 214850

**POSTERIOR**
Screenshot 2026-01-27 214938

### DIFF
--- a/packages/niivue/src/niivue/index.ts
+++ b/packages/niivue/src/niivue/index.ts
@@ -3646,7 +3646,7 @@ export class Niivue {
         this.drawScene()
     }
 
-   /**
+    /**
      * Snap the 3D render camera to a specific anatomical view.
      * @param view - The target view. Supported values (case-insensitive): 'Top'/'Superior', 'Bottom'/'Inferior', 'Left', 'Right', 'Front'/'Anterior', 'Back'/'Posterior'.
      * @see {@link https://github.com/niivue/niivue/issues/1464 | Issue 1464}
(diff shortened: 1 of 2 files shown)


AviJxn:
Thanks for the clarification @neurolabusc !

I understand the distinction between the core library and the desktop implementation now. It was a great learning experience digging into the camera logic and setting up the unit tests regardless.

I'll leave this branch here in case @cdrake wants to reference the renderAzimuth/Elevation logic for the desktop implementation later. Thanks for the review!

PR #1547, opened 2026-02-12
Support tensor brightness scaling using `cal_min` and `cal_max` properties

kabilar:
- This pull request adds the ability to modify the brightness of a tensor image that is displayed with `loadImgV1` using the `cal_min` and `cal_max` properties.
- The default behavior relies on the existing implementation of utilizing the image's maximum value.
- I tested this locally using the [modulateAfniDTI](https://github.com/niivue/niivue/blob/main/packages/niivue/demos/features/modulateAfniDTI.html) demo.  As I am new to NiiVue development, I would appreciate any further testing that you can do.  Thank you.

cc @satra @ayendiki

### DIFF
--- a/packages/niivue/src/niivue/index.ts
+++ b/packages/niivue/src/niivue/index.ts
@@ -6362,6 +6362,10 @@ export class Niivue {
         this.refreshColormaps()
         this.closePAQD()
         for (let i = 0; i < numLayers; i++) {
+            // refresh V1 tensor data with current cal_min/cal_max
+            if (this.volumes[i].v1) {
+                this.volumes[i].loadImgV1()
+            }
             // avoid trying to refresh a volume that isn't ready
             if (!this.volumes[i].toRAS) {
                 continue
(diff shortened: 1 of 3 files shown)


neurolabusc:
@kabilar while direct PRs are fine, most NiiVue changes begin life as an issue, where one provides a concrete example that illustrates the current issue and prompts discussion regarding the optimal solution. I always like making a /niivue/packages/niivue/src/index.html file that illustrates the issue, so others can explore and resolve it with a hot-reloadable demo using `npm run dev`. So it would be great to adapt [modulateAfniDTI](https://niivue.com/demos/features/modulateAfniDTI.html) to illustrate the problem you are trying to solve (which is still not clear to me).

For your data, it might be worth looking at the [timeseries2 live demo](https://niivue.com/demos/features/timeseries2.html) that illustrates changing the calMinMax on a per-frame basis (we tend to use the term `frame` for 4D timeseries).

kabilar:
Closing due to the issue that @neurolabusc mentioned in https://github.com/niivue/niivue/issues/1548.  Looking into an alternative solution.

Issue #1556, opened 2026-02-23
FEATURE: Adding manual pivot point


dekesh:
I’m currently using NiiVue to visualize a mesh (sEEG contacts) together with the corresponding MRI. In this workflow, it would be very useful to focus on a specific contact and interactively zoom/scale, and rotate around that exact point.

At the moment, this is not possible because the pivot point is computed automatically and cannot be overridden.

I would like to propose a minimal PR that introduces an API option to explicitly set the pivot point for camera transformations. This would allow users to set/get the center of rotation/zoom.

If there is interest in this feature, I’d be happy to open a small PR implementing this functionality.

Demo attached:

https://github.com/user-attachments/assets/2cc1d107-34ce-4e26-8975-d83ce23f41ca

PR #1557, opened 2026-02-23
Feature: Adding manual pivot point

dekesh:
This is a minimal PR based on #1556 

It adds a basic API so that user can set the pivot point  (in the `mm` coordinate system), and adds a demo page to help potential user to user the API

API:
------
`setPivot3DPoint(pivot3DPoint: [number, number, number] | null)` : set the pivot point to required value or potentially set to null to restore automatic behavior. Silently ignore NaN or Infinite values,

`get pivot3D()`: Get the current pivot point (this existed before as a public variable, but could not be written by the user in any case, as the pivot was automatically calculated.

The reason for the mismatch between the `setPivot3D..` and the `pivot3D` is because there is already a public `setPivot3D` point (which does the heavy lifting), and I didn't want to break anything: If that is not a good enough solution, I would be happy to change as needed.

Implementation logic:
--------------------------------------

* Allow the user to set a desired pivot point `setPivot3DPoint(pivot3DPoint)`. Store in a dedicated private variable (`pivot3DUser`), and draw scene.
* `setPivot3D`: Check if the user set a (valid) pivot point, and use it. Make sure that all scene is generated by setting `furthestFromPivot` correctly (furthest corner from pivot point).

Demo page:

https://github.com/user-attachments/assets/0af77b37-b5a5-467e-b749-8d020ed4da4c

### DIFF
--- a/packages/niivue/demos/features/camera.pivot.html
+++ b/packages/niivue/demos/features/camera.pivot.html
@@ -50,16 +50,12 @@
   const pzEl = document.getElementById("pz")
 
   function updateStatus() {
-    const pivot = nv.getPivot3DPoint()
-    if (pivot) {
-      status.textContent = `Current pivot: [${pivot[0].toFixed(2)}, ${pivot[1].toFixed(2)}, ${pivot[2].toFixed(2)}]`
-    } else {
-      status.textContent = "Current pivot: auto"
-    }
+    const pivot = nv.pivot3D
+    status.textContent = `Current pivot: [${pivot[0].toFixed(2)}, ${pivot[1].toFixed(2)}, ${pivot[2].toFixed(2)}]`
   }
 
   function updateInputsFromPivot() {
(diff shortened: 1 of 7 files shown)


dekesh:
Just updating that tests seems to work locally (but some seems flaky), will get back to that soon

dekesh:
@hanayik : Do you happen to have any insight about this failure in the test? As far as I understand the PR has nothing to do with this. On the other hand it seems to work for other PRs.

The test fails because it reach the `30 sec` limit of a single playwright test.
From local debugging (locally the test pass after about  10 seconds), the most time-intensive parts in the `evaluate` block are
the `setMeshProperty` Commands. I've added some timing to check that on the CI. The tests seems to be passing even when the `setMeshPropery` lines are missing. So removing these may shave  some seconds..

dekesh:
Another day I guess :)

PR #1563, opened 2026-02-26
gradient fixes

thewtex:
- **fix: render gradient texture for all slices in gradientGL**
- **fix: guard against zero-length gradient in normalize() calls**


thewtex:
@neurolabusc @hanayik please take a look. A few bugs in the gradient shaders. Addressing the issue Taylor brought up https://github.com/fideus-labs/fidnii/pull/71#issuecomment-3960505698

image

hanayik (approved):

neurolabusc:
@thewtex the change to the compute shader fixes a clear bug. However, note that the amplitude of the gradient is written in the alpha component of the RGBA. Therefore, my own sense is the way to use GLSL would be to use the precomputed length instead of recomputing it per fragment. I do think that length is highly optimized on GPUs, but this is also in the inner loop of the raycast pipeline. So my sense preference would be:

```
vec4 grad = texture(gradient, samplePos.xyz);
if (grad.a > 0.0) {
   grad.rgb = normalize(grad.rgb*2.0 - 1.0);
}
```

thewtex:
@neurolabusc good catch! I added a commit accordingly.

Issue #1564, opened 2026-02-26
Single 2d view doesn't fill entire canvas


pwighton:
When viewing a single 2d view (Axial, Coronal, or Sagittal), the viewport is locked to the volumes's aspect ratio, which creates empty space on the canvas when the user zooms in.

To reproduce:

- Navigate to: https://niivue.com/demos/features/ui.html
- Select 'View -> Axial' and 'Drag -> Pan'
- Use the scroll wheel to zoom in

<img width="1902" height="1013" alt="Image" src="https://github.com/user-attachments/assets/d47b0aef-8fe0-45b1-b634-a07e1a3af96c" />

pwighton:
Resolved with #1595

PR #1565, opened 2026-02-26
Make isStretchToScreen pathway reachable to fix 2d view not filling entire canvas issue

pwighton:

- #1564

This is a proposed fix to issue #1564.  However I'm not sure if this fix would have other unintended consequences.

## Cause 

1. The [single-panel draw path in `index.ts`](https://github.com/niivue/niivue/blob/f05bc78ee460f8432345bc9aa5066014b6f03d63/packages/niivue/src/niivue/index.ts#L11864) computes aspect-ratio-constrained pixel dimensions and passes them to `draw2D()`:

2. [`draw2D()`](https://github.com/niivue/niivue/blob/f05bc78ee460f8432345bc9aa5066014b6f03d63/packages/niivue/src/niivue/index.ts#L9744) uses those dimensions to shrink the viewport and center it with padding, before calling `draw2DMain()`. However, `draw2DMain()` has an [`isStretchToScreen` path](https://github.com/niivue/niivue/blob/f05bc78ee460f8432345bc9aa5066014b6f03d63/packages/niivue/src/niivue/index.ts#L9544) designed to fill the canvas by expanding the mm-space FoV to match the canvas aspect ratio, but it's unreachable. It triggers on `leftTopWidthHeight[2] === 0 || leftTopWidthHeight[3] === 0`, and `draw2D()` always resolves `[0,0,0,0]` to concrete canvas dimensions before the call.

## Fix

1. Single-panel draw path: Stop computing `actualDimensions` and pass `[0,0,0,0]` through without constraining the viewport:

2. In `draw2D()`, when `imageWidthHeight` is `NaN` and the original input was `[0,0,0,0]`, preserve it for `draw2DMain()` so `isStretchToScreen` triggers.

## Result

Before:

niivue-zoom-issue-before

After:

niivue-zoom-issue-after

### DIFF
--- a/packages/niivue/src/niivue/index.ts
+++ b/packages/niivue/src/niivue/index.ts
@@ -9740,7 +9740,12 @@ if (perm[0] === 1 && perm[1] === 2 && perm[2] === 3) {
         }
 
         if (isNaN(imageWidthHeight[0])) {
-            this.draw2DMain(ltwh, axCorSag, customMM)
+            // If original input had zero w/h (single-panel full-screen draw),
+            // pass [0,0,0,0] to draw2DMain to trigger isStretchToScreen,
+            // which expands the mm-space FoV to fill the canvas while
+            // keeping square pixels. This allows zoom to use the full canvas.
+            const wasZeroSized = leftTopWidthHeight[2] === 0 && leftTopWidthHeight[3] === 0
+            this.draw2DMain(wasZeroSized ? [0, 0, 0, 0] : ltwh, axCorSag, customMM)
         } else {
             // inset as padded in tile
             // issue1554 shift tile: do not change global padLeftTop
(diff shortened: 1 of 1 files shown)


aarmea:
Do we have any updates on this? This fix seems like it would genuinely improve the UX for our use case.

@pwighton if you don't currently have cycles, I don't mind taking over the last mile of this regarding adding a test and otherwise looking for regressions.

pwighton:
Thanks for offering to help, @aarmea.  I am overloaded at the moment and would happily welcome assistance to get this merged! :smile:

hanayik:
@aarmea , feel free to carry it over the finish line. I'm sure @pwighton wouldn't mind. Would be happy to merge soon and release a new version after your updates.

aarmea:
I have the Playwright tests in my fork here: https://github.com/aarmea/niivue/tree/u/aarmea/1564/2d-fill-canvas, which includes both positive tests that ensure that the single-pane mode works correctly, but also negative tests that ensure that multiplane views (like the default) are not affected. I'll follow up with how manual testing of the demos goes after I have the chance to go through them.

Should I open a new PR, or would you rather pull the changes in here instead?

hanayik:
@aarmea , a new PR would be fine (and probably a bit cleaner to deal with). Thanks for getting it over the finish line!

aarmea:
Sounds good, I'll go ahead and make one then -- see #1595. A cursory manual testing pass shows no regressions.

PR #1572, opened 2026-03-06
fix: emit azimuthElevationChange event during mouse drag rotation

thewtex:
mouseMove() directly set scene.renderAzimuth/renderElevation and called
drawScene(), bypassing setRenderAzimuthElevation() which is the only
place that emits the azimuthElevationChange event. This meant external
listeners (e.g. for syncing two NiiVue instances) were never notified
of camera rotation caused by mouse drag — only keyboard arrow keys
triggered the event.

Replace the three direct lines with a single call to
setRenderAzimuthElevation(), which atomically sets both values, emits
the event, and calls drawScene(). This is consistent with how keyboard
rotation already works.

### DIFF
--- a/packages/niivue/src/niivue/index.ts
+++ b/packages/niivue/src/niivue/index.ts
@@ -1,130 +1,132 @@
 import { mat4, vec2, vec3, vec4 } from 'gl-matrix'
+
 import packageJson from '../../package.json' with { type: 'json' }
 import { TEXTURE_CONSTANTS } from './texture-constants'
-import { orientCube } from '@/orientCube'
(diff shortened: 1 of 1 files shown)


thewtex:
I ran `lint:fix`, and I am not sure why there are so many changes.

cdrake:
Sorry for the delay.  I will review this tomorrow.

cdrake:
Can you check if you have a linter configured in your IDE? Please undo the lint and I will review the PR.

thewtex:
@cdrake thank you! I re-added in #1573

Issue #1578, opened 2026-03-21
Bug displaying colormaps with more than 256 colors


BorisMansencal:
I'm new to Niivue, and I find it really impressive. Great work!

I'm trying to display an atlas on a T1w image. Niivue seems perfect for this.

However, there seems to be a bug when the atlas has more than 256 colors.

Here's an example: https://dept-info.labri.fr/~mansenca/BUGNIIVUE/test0.html where I'm displaying an atlas with 353 labels onto itself.

The background color should be transparent, but it's green instead (the color of label index 2).

## Environment

- NiiVue version: 0.68.1
- Browser: Chrome, Brave, Firefox
- OS: macOS

BorisMansencal:
**According to Claude Code**, this problem is related to the fix for issue #1139 in the package/niivue/src/shader-srcs.ts file:
```typescript
//issue1139: survives threshold, so round up to opaque voxel
txl = max(txl, 2.0/256.0);
```
This fix seems to assume a 256-entry colormap texture. When a `colormapLabel` is used with more than ~256 entries, this clamp shifts the texture lookup for low voxel values (0, 1, 2...) to the wrong label index.

This clamp was added for standard 256-entry colormaps to skip reserved transparent entries at positions 0–1. However, when `setupColormapLabel()` (in `src/niivue/data/VolumeColormap.ts`) creates a label colormap texture, the texture has `nLabel` entries (e.g. 354), not 256. The `2.0/256.0` threshold then maps to a wrong pixel:

**For voxel value 0 with a 354-label colormap:**
1. `setupColormapLabel` sets `cal_min = -0.5`, `cal_max = 353.5`
2. Shader computes: `txl = (0 - (-0.5)) / 354 = 0.00141` — correctly points to texture index 0
3. But `f (0) > mn (-0.5)` is **true**, so the clamp fires: `txl = max(0.00141, 2.0/256.0) = 0.00781`
4. In a 354-pixel texture: `0.00781 × 354 ≈ 2.76` → **looks up index 2 instead of index 0**

The same clamp appears a second time at line ~1321 for the negative color path.

## Proposed fix

Add a `uniform bool isColormapLabel` to the shader and skip the `2.0/256.0` clamp when rendering label colormaps. Label colormaps do not have reserved transparent entries at positions 0–1, so the clamp should not apply.

### 1. `src/shader-srcs.ts` — `fragOrientShader`

Add the uniform declaration:
```glsl
uniform bool isColormapLabel;
```

[...]

PR #1579, opened 2026-03-21
Fix to handle label colormaps with more than 256 entries

BorisMansencal:
Fix to issue #1578 proposed by Claude Code.

## Summary

In `fragOrientShader`, the hardcoded clamp `txl = max(txl, 2.0/256.0)` (introduced for issue #1139) assumes a 256-entry colormap texture. When a `colormapLabel` is used with more than ~256 entries, this clamp shifts the texture lookup for low voxel values (0, 1, 2...) to the wrong label index. In practice, **label 0 (background) is never looked up** — a nearby label's color is displayed instead.

## Steps to reproduce

1. Load a base volume (e.g. a T1w image) and an atlas overlay with integer labels 0–353
2. Define a `colormapLabel` where:
   - Index 0 = transparent background `[R=0, G=0, B=0, A=0]`
   - Index 1 = red `[R=255, G=0, B=0, A=255]`
   - Index 2 = green `[R=0, G=255, B=0, A=255]`
3. Apply the colormap with `volume.setColormapLabel(cmap)` and call `nv.updateGLVolume()`

## Expected behavior

Voxels with value 0 should be **transparent**, showing the base T1w image underneath.

## Actual behavior

Voxels with value 0 appear **green** (the color of label index 2). Changing the color assigned to index 0 has no visible effect — the shader never reads that entry.

## Root cause

In `src/shader-srcs.ts`, the `fragOrientShader` contains (lines ~1301–1303):

```glsl
float txl = mix(0.0, 1.0, (f - mn) / r);
if (f > mn) { //issue1139: survives threshold, so round up to opaque voxel
    txl = max(txl, 2.0/256.0);
}
```

This clamp was added for standard 256-entry colormaps to skip reserved transparent entries at positions 0–1. However, when `setupColormapLabel()` (in `src/niivue/data/VolumeColormap.ts`) creates a label colormap texture, the texture has `nLabel` entries (e.g. 354), not 256. The `2.0/256.0` threshold then maps to a wrong pixel:

**For voxel value 0 with a 354-label colormap:**
1. `setupColormapLabel` sets `cal_min = -0.5`, `cal_max = 353.5`
2. Shader computes: `txl = (0 - (-0.5)) / 354 = 0.00141` — correctly points to texture index 0
3. But `f (0) > mn (-0.5)` is **true**, so the clamp fires: `txl = max(0.00141, 2.0/256.0) = 0.00781`
4. In a 354-pixel texture: `0.00781 × 354 ≈ 2.76` → **looks up index 2 instead of index 0**

The same clamp appears a second time at line ~1321 for the negative color path.

## Impact

- **Label 0 is unreachable**: the shader can never look up the first ~2 entries of a label colormap with >256 entries. Background voxels display a wrong (opaque) color instead of being transparent.
- **Labels 1–2 are also affected**: they are clamped to the same texture coordinate as label 0, so they all display the color of label ~2–3.
- The bug is **silent** — no console errors, just incorrect rendering.

## Proposed fix

Add a `uniform bool isColormapLabel` to the shader and skip the `2.0/256.0` clamp when rendering label colormaps. Label colormaps do not have reserved transparent entries at positions 0–1, so the clamp should not apply.

### 1. `src/shader-srcs.ts` — `fragOrientShader`

Add the uniform declaration:
```glsl
uniform bool isColormapLabel;
```

Guard both clamp occurrences:
```glsl
// Line ~1302 (positive colors)
if (f > mn && !isColormapLabel) {
    txl = max(txl, 2.0/256.0);
}

// Line ~1321 (negative colors)
if (!isColormapLabel) {
    txl = max(txl, 2.0/256.0);
}
```

### 2. `src/niivue/data/VolumeColormap.ts` — `setupColormapLabel()`

Set the uniform to `1` when a label colormap is active:
```typescript
gl.uniform1i(orientShader.uniforms.isColormapLabel, 1)
```

### 3. `src/niivue/index.ts` — regular colormap path

Reset the uniform to `0` when using a standard colormap (when `colormapLabelTexture === null`):
```typescript
this.gl.uniform1i(orientShader.uniforms.isColormapLabel, 0)
```

### DIFF
--- a/packages/niivue/src/niivue/data/VolumeColormap.ts
+++ b/packages/niivue/src/niivue/data/VolumeColormap.ts
@@ -78,6 +78,7 @@ export function setupColormapLabel(params: ColormapLabelParams): ColormapLabelRe
 
     gl.uniform1f(orientShader.uniforms.cal_min, overlayItem.colormapLabel.min! - 0.5)
     gl.uniform1f(orientShader.uniforms.cal_max, overlayItem.colormapLabel.max! + 0.5)
+    gl.uniform1i(orientShader.uniforms.isColormapLabel, 1)
     gl.bindTexture(gl.TEXTURE_2D, colormapLabelTexture)
 
     return {
--- a/packages/niivue/src/niivue/index.ts
+++ b/packages/niivue/src/niivue/index.ts
@@ -6932,6 +6932,7 @@ if (perm[0] === 1 && perm[1] === 2 && perm[2] === 3) {
             createColormapTexture: this.createColormapTexture.bind(this)
         })
         if (colormapLabelTexture === null) {
+            this.gl.uniform1i(orientShader.uniforms.isColormapLabel, 0)
             this.gl.bindTexture(this.gl.TEXTURE_2D, this.colormapTexture)
             this.gl.uniform1f(orientShader.uniforms.cal_min, overlayItem.cal_min!)
             this.gl.uniform1f(orientShader.uniforms.cal_max, overlayItem.cal_max!)
--- a/packages/niivue/src/shader-srcs.ts
+++ b/packages/niivue/src/shader-srcs.ts
@@ -1276,6 +1276,7 @@ uniform float cal_minNeg;
 uniform bool isAlphaThreshold;
 uniform bool isColorbarFromZero;
 uniform bool isAdditiveBlend;
+uniform bool isColormapLabel;
 uniform highp sampler2D colormap;
 uniform lowp sampler3D blend3D;
 uniform int modulation;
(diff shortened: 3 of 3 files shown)


neurolabusc:
Note that indexed atlases use the [fragOrientShaderAtlas](https://github.com/niivue/niivue/blob/c1e44022b5bd1e103b69f117cab46419ff637fce/packages/niivue/src/shader-srcs.ts#L1173) instead of the `fragOrientShader`. An atlas has discrete regions, rather than scalar values so requires a nearest neighbor interpolation: a voxel that is a mixture of area 17 and 19 should not be given the number 18. 

A nice example of an Atas having more than 256 regions is the [sparse INRIA19 atlas live demo](https://niivue.com/demos/features/atlas.sparse.html) where the largest index is [1606](https://github.com/niivue/niivue/blob/c1e44022b5bd1e103b69f117cab46419ff637fce/packages/niivue/demos/images/inia19-NeuroMaps.json#L6). Notice how that live demo uses `setColormapLabelFromUrl`. 

A well behaved [NIfTI header](https://brainder.org/2012/09/23/the-nifti-file-format/) should have the proper intent code (NIFTI_INTENT_LABEL = 1002).

PR #1583, opened 2026-04-04
Fix Otsu thresholding on non-RAS images

aarmea:
Previously, when `Niivue.drawOtsu()` applied the threshold to the bitmap, it did so atop the voxel data as it exists in the source file regardless of any voxel-space transforms Niivue may be performing after loading. This change edits `drawOtsu()` to use the transformed image as returned by `img2RAS()` rather than using the raw image from the volume.

## Screenshots

I obtained these on the ["Drawing User Interface" demo](https://niivue.com/demos/features/draw.ui.html) with the LAS image already in this repository at `tests/images/nifti_space/LAS.nii.gz`.

### Before

Note the incorrect highlight on the LR:

image

### After

The LR is fixed:

image

## Future work

My need (and ability to test) was only for `drawOtsu()`, but on the surface it seems some of the other functions, in particular `removeHaze()` might need a similar fix.

### DIFF
--- a/packages/niivue/src/niivue/index.ts
+++ b/packages/niivue/src/niivue/index.ts
@@ -3535,7 +3535,7 @@ export class Niivue extends EventTarget {
             this.createEmptyDrawing()
         }
         const result = ImageProcessing.applyOtsuToDrawing({
-            img: this.volumes[0].img!,
+            img: this.volumes[0].img2RAS(),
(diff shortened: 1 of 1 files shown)


hanayik:
@aarmea thanks for the fix. I will get this merged soon and also make sure this issue is not present in the upcoming Niivue re-write (in a separate repo).

hanayik (approved):

PR #1591, opened 2026-04-20
Feature: Smooth bitmaps and global clipping

aarmea:
# Description

This PR implements two optional, related features that improve the usability of bitmaps, particularly when they are used for segmentation or to indicate the boundary of a volume:
* Smoothed rendering of bitmap surfaces
* Clip planes that clip bitmaps and all volumes, not just the first one

## Screenshots

These were taken with the "Drawing User Interface" demo page with the volume it loads by default. I applied an Otsu with 2 segmentation classes and switched the view to "Render" only.

### Before

This is Niivue's current behavior in this scenario (and continues to be the default after this change).

Note how the jagged/blocky appearance of the bitmap makes it difficult to see the border of the volume or inside it.

niivue-before

### With surface smoothing

Surface smoothing is enabled by setting `nv.opts.smoothDrawing` to a nonzero value. In this example, it is set to 2.0. This can be done in the demo page at Draw --> Smooth Drawing 3D.

Note how the smoothing makes it possible to see the surface of the gray matter, even where the volume is not rendered because of the location of the clip plane.

niivue-smooth-only

### With surface smoothing and global clipping

Global clipping is enabled by setting `nv.opts.isClipAllVolumes = true`. This can be done in the demo page at View --> Clip All Volumes.

Note how the clipping makes the clip plane even more visible. It effectively cuts a hole into the bitmap, allowing the user to look inside to see the volume.

niivue-smooth-clip

Users who are familiar with the Hyperfine Swoop system user interface will recognize this as the "3D 2" view mode.

### Smooth bitmap surface rendering

This is implemented in three phases:
1. A blurred version of the bitmap is created on the CPU. To do this efficiently, it is implemented as box blurs instead of a Gaussian blur. To avoid unresponsiveness, this is done on a Web worker (much like Brainchop).
2. The blurred bitmap is sent to the GPU as a R16F 3D texture with LINEAR filtering for additional free smoothing.
3. The render shader uses the blurred bitmap texture to find the isosurface and calculate the surface normal, which it then uses to apply Phong shading.

### Global clip planes

When rendering volumes and `isClipAllVolumes` is set, the shader simply skips over voxels that are behind the clip plane, regardless of which volume it is rendering.

## Generative AI disclosure

I used GitHub Copilot with Claude Sonnet and Opus 4.6 as a coding agent to assist with this change.

## Funding

This change was funded by [Hyperfine, Inc.](https://www.hyperfinemri.com) and is released under the BSD 2-Clause License.


neurolabusc:
@aarma this is a really interesting idea—thanks for putting it together. It might be helpful to start this as an issue so we can get broader community input on the use case, potential overlap, and implementation options.

There does seem to be some overlap with the [setRenderDrawAmbientOcclusion() option](https://niivue.com/demos/features/torso.html), so it would be good to clarify how this extends or differs from that behavior.

Since the blurred data ultimately needs to live on the GPU, this could be a nice opportunity to leverage GLSL shaders rather than round-tripping through the CPU. One possible starting point is the [blurFragShader() fragment shader](https://github.com/niivue/niivue/blob/154fb6c7bd966f1abcb41c1ecf39a4b073159368/packages/niivue/src/shader-srcs.ts#L2210), which might fit well with what you’re proposing.

aarmea:
@neurolabusc thanks for the quick feedback.

> There does seem to be some overlap with the [setRenderDrawAmbientOcclusion() option](https://niivue.com/demos/features/torso.html), so it would be good to clarify how this extends or differs from that behavior.

Going through the linked example, it seems that our approaches are doing completely different things. In particular, I couldn't find a value for occlusion that removes the blockiness, and the lack of surface normals and shading makes it difficult to see the geometry.

| Occlusion at minimum | Occlusion at maximum |
|-----|-----|
| image    |  image |

I patched the torso demo to use my smooth bitmaps feature, and while it does perform some of the smoothing, it also revealed a bug that I missed in my testing with other Niivue and company internal volumes which will need to be fixed regardless:

image

> Since the blurred data ultimately needs to live on the GPU, this could be a nice opportunity to leverage GLSL shaders rather than round-tripping through the CPU. One possible starting point is the [blurFragShader() fragment shader](https://github.com/niivue/niivue/blob/154fb6c7bd966f1abcb41c1ecf39a4b073159368/packages/niivue/src/shader-srcs.ts#L2210), which might fit well with what you’re proposing.

That's a great idea that could improve the performance of the feature even more. Hopefully it won't be too much of a lift to implement -- since I have the aforementioned bug to fix, it's an excuse to try it out, and maybe it'll even be performant enough to implement the Gaussian blur instead of the box blurs.

aarmea-hyperfine:
Here's what I just pushed up:
* The blur is now performed via GLSL -- it's actually using the same `blurFragShader` you mentioned, which I modified slightly to allow a window larger than 3x3 if specified. (I would have preferred to call it once per draw both here and in `gradientGL` rather than once per slice, but it seems there isn't a way to call a shader with a complete 3D texture in WebGL 2...)
* I slightly optimized the render shader when bitmap smoothing is enabled.
* I fixed the bug I mentioned above, where we were sometimes using the wrong bitmap color:

image

hanayik:
@aarmea , I'll merge and release a new version once you have added a new playwright test. You'll need to generate the first snapshot and commit it so that the github actions runner has something to compare against.

aarmea:
Tests are up. Thanks everyone for the quick turnaround and feedback on this.

hanayik (approved):
Looks good. My final comment is that when viewing in multiplanar + 3D render mode, drawings are normally visible in the volume render tile, even when the mask is only occupying a single slice. 

When `smoothDrawing > 0`, drawings may not appear until enough adjacent slices also have data to smooth with. This isn't a blocker, and since all the settings are "opt-in" I think it's ok as is. It is just a slight difference in behaviour due to the smoothing operation. That's not really avoidable. 

## No smoothing
Image

## With smoothing
Image

PR #1609, opened 2026-05-14
Fix NPY header parsing and dtype handling

elebot4:
### Summary 

This PR fixes several issues with the existing `.npy` reader:

- handling different types of npy version (v1/2/3) and the different headers is now done correctly
- now passes correctly the endian of the .npy file to the created nvImage object 
- now properly reads values for 4d arrays of shape [c, x, y, z]

### Motivations 

Some valid npy files were being read incorrectly. In particular, signed integer label maps and 4D arrays could be interpreted with incorrect metadata.

### Tested

Manually tested through the VS Code viewer using a local NiiVue build. 
- 3D int16 label array with values `-1..3`
- 4D float32 image array with shape `(4, 162, 192, 149)`, displayed as matrix size `149 × 192 × 162` with `4` timepoints

### Notes

A similar issue may exist in the new `niivue/mono` repository. I opened this PR here because `niivue-vscode` currently only depends on this package.

### DIFF
--- a/packages/niivue/src/nvimage/ImageReaders/numpy.ts
+++ b/packages/niivue/src/nvimage/ImageReaders/numpy.ts
@@ -8,18 +8,30 @@ import { Zip } from '@/nvutilities'
  * @returns Byte size of the data type
  */
 function getTypeSize(dtype: string): number {
-    const typeMap: Record<string, number> = {
-        '|b1': 1, // Boolean
-        '<i1': 1, // Int8
-        '<u1': 1, // UInt8
-        '<i2': 2, // Int16
-        '<u2': 2, // UInt16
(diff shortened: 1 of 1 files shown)


hanayik:
@elebot4 , could you fix the formatting issues and push the the changes please? Otherwise, looks ok. It would be ideal to have a test case for this too (with very small npy arrays that cover the scope of your manual checks).

elebot4:
Thanks for reviewing. Sorry about the formatting issues. I ran ESLint and pushed the changes.

I haven’t added the .npy test case. Adding a Python/NumPy script to generate the arrays would introduce a new test dependency, while hand-written binary fixtures seem fragile and not very exhaustive.

PR #1610, opened 2026-05-15
Niivue build/CI: restore eslint plugins and add SwiftShader test slack

cdrake:
## Summary

Two small build/CI fixes for the `@niivue/niivue` package, extracted from the desktop work on `BIDS-allineate` so they can land on `main` independently.

- **Restore `eslint-plugin-{import,n,promise}` as niivue devDependencies.** `eslint-config-standard` pulls these in as peer dependencies; without them declared locally, `npm ci` on a clean tree (CI) can't resolve the config and lint fails with *"ESLint couldn't find the plugin eslint-plugin-import"*. The plugins were silently dropped from the lockfile during an earlier `npm install` and never re-added.
- **Allow SwiftShader rendering slack in the depthtest regression spec.** Mesh + volume rendering through GitHub Actions' SwiftShader software rasterizer is significantly slower and produces noticeably different anti-aliased output than a real GPU. The test now sets a 90 s wall-clock timeout and a 25% `maxDiffPixelRatio` so software-rendered tractography lines don't cause a flaky CI mismatch, while still catching a mesh that's entirely missing or hidden behind the volume.

No `packages/niivue/src/` changes; this is purely build tooling + test tolerance.

## Test plan

- [ ] `npm ci` on a clean tree resolves without the `eslint-plugin-import` error.
- [ ] `npm run lint -w @niivue/niivue` exits 0 (warnings allowed; the pre-existing `max-lines-per-function` warnings are unchanged).
- [ ] `npm run test-playwright -w @niivue/niivue -- --grep "depth test"` passes locally and on the SwiftShader runner in CI.


cdrake:
@hanayik - This PR encompasses the few changes in NiiVue to accommodate the large desktop workflow PR.

hanayik (approved):

PR #1614, opened 2026-06-11
perf: cache the raw volume texture across refreshLayers calls

pFornagiel:
# perf: cache the raw volume texture across refreshLayers calls

`updateGLVolume()` runs on many display-only interactions (window/level, opacity, colormap changes). Each call rebuilt and re-uploaded the full raw 3D volume texture even though the voxel data hadn't changed. For large volumes, and float64 volumes in particular, that means a full `Float32Array.from(...)` conversion plus a large CPU-to-GPU upload on every slider tick, which causes visible stalls.

This PR reuses the already-uploaded raw texture when the source voxel data is unchanged, so display-only updates only re-run the orient pass.

Not going to hide that this change was made using help of Claude, but it did help with the performance of our personal project, so I figured it may be worth a review.

## Changes

- `Niivue` keeps a per-layer cache of the raw 3D texture, keyed by the source `img` array reference and `frame4D`.
- In `refreshLayers()`, if the cache entry matches (same `img` reference, same `frame4D`, datatype is not `DT_RGBA32`), the existing raw texture is reused and the conversion/upload is skipped. On a miss, the previous upload path runs and the cache entry is replaced.
- Orient-shader selection was extracted into `VolumeLayerRenderer.selectVolumeOrientShader()` so it can run on cache hits without going through the upload path.
- Cached textures are no longer unconditionally deleted at the end of `refreshLayers()`. Stale entries are deleted when a layer is re-uploaded, and entries for removed layers are swept in `updateGLVolume()`.
- `DT_RGBA32` is excluded from reuse because the PAQD handling has side effects in the upload path that must still run.

Display parameters (`cal_min`/`cal_max`, opacity, colormap) are applied as orient-pass uniforms, so they stay correct while the raw upload is skipped. The orient pass and everything downstream are unchanged.

## Benchmarks

Measured with an interleaved A/B harness (Chromium 143 headless, i5-8365U / UHD 620, Linux) on a float64-heavy CBCT dataset (ToothFairy3). Baseline is `main` at a04235c8, candidate is this branch. Final-frame checksums are identical between baseline and candidate in every scenario.

| Scenario | Volume | Baseline median | This PR | Speedup |
|---|---|---:|---:|---:|
| cold_load | small | 1249 ms | 1200 ms | 1.04x |
| cold_load | medium | 2536 ms | 2439 ms | 1.04x |
| cold_load | large | 6796 ms | 6615 ms | 1.03x |
| intensity_sweep | small | 319 ms | 190 ms | 1.68x |
| intensity_sweep | medium | 657 ms | 473 ms | 1.39x |
| intensity_sweep | large | 1489 ms | 1003 ms | 1.49x |
| opacity_sweep | medium | 604 ms | 412 ms | 1.47x |
| colormap_cycle | medium | 1389 ms | 1230 ms | 1.13x |
| overlay_intensity_sweep | overlay | 301 ms | 214 ms | 1.41x |

The wins are concentrated where expected: repeated display-only updates run 1.4–1.7x faster, while cold load is near-neutral since the first upload still has to happen.

## Trade-offs

- VRAM usage goes up: one raw texture per cached layer stays resident between refreshes.
- Invalidation is identity-based. Mutating `img` in place without replacing the array reference will keep serving the stale texture. This matches how NiiVue replaces volume data today, but it's a constraint worth knowing.
- Not benchmarked: 4D frame-switch-heavy workloads, the `DT_RGBA32`/PAQD path (intentionally bypassed), and datatypes beyond the float64-dominant dataset.

A reasonable follow-up would be unit/integration coverage for cache invalidation around `frame4D` changes and layer removal.

### DIFF
--- a/packages/niivue/src/niivue/data/VolumeLayerRenderer.ts
+++ b/packages/niivue/src/niivue/data/VolumeLayerRenderer.ts
@@ -474,13 +474,28 @@ export interface SetupVolumeTextureDataResult {
 }
 
 /**
- * Setup volume texture data based on datatype.
- * Allocates GPU texture storage, transfers image data, and selects appropriate shader.
- * @param params - Setup parameters
+ * Parameters for selecting the orient shader for non-RGBA32 volume data
+ */
+export interface SelectVolumeOrientShaderParams {
+    gl: WebGL2RenderingContext
+    hdr: any
+    orientShaderU: Shader
+    orientShaderI: Shader
+    orientShaderF: Shader
+    orientShaderRGBU: Shader
+    orientShaderAtlasU: Shader
+    orientShaderAtlasI: Shader
+}
+
+/**
+ * Select the orient shader matching the volume datatype and intent code.
+ * Shared by setupVolumeTextureData() and the cached-texture path in
+ * refreshLayers(), which skips the GPU upload but still needs the shader.
+ * @param params - Shader selection parameters
  * @returns Selected orient shader
  */
-export function setupVolumeTextureData(params: SetupVolumeTextureDataParams): SetupVolumeTextureDataResult {
-    const { gl, hdr, img, orientShaderU, orientShaderI, orientShaderF, orientShaderRGBU, orientShaderAtlasU, orientShaderAtlasI } = params
+export function selectVolumeOrientShader(params: SelectVolumeOrientShaderParams): Shader {
+    const { gl, hdr, orientShaderU, orientShaderI, orientShaderF, orientShaderRGBU, orientShaderAtlasU, orientShaderAtlasI } = params
 
     let orientShader = orientShaderU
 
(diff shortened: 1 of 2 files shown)


neurolabusc:
@pFornagiel, as noted from the project’s front page, this project is the legacy beta release of NiiVue. We’ve moved the active development to the [mono](https://github.com/niivue/mono/issues/48) repository, which is where the 1.0 release candidates are coming from. The 1.0 API now uses a model-view-controller pattern, so your adapter only talks to the controller. This change lets us use multiple backend views (like WebGL2 and WebGPU) without any issues. Plus, the controller can handle batching automatically, which means the user does not have to explicitly call updateGLVolume() to invoke deferred expensive calculations

We’ll keep the pre-1.0 versions available for users who are still using our legacy codebase, but we won’t be adding any new features.

By default, we use the WebGPU backend, but we’ll switch to WebGL2 if WebGPU isn’t supported. The new [performance benchmarks](https://niivue.github.io/mono/examples/) are a great way to see how our new architecture works.
mvc


This is an open-source project with a permissive license, so feel free to create your own version of the legacy tool. However, we’re a small team, and we need to focus our efforts on the refactored code. As you mentioned in your limitations, the way our legacy code works was a conscious decision based on the trade-offs, so I don’t think we should change it for those who rely on it. But I do believe that our fresh approach with the new API really addresses your concerns.

pFornagiel:
Hi @neurolabusc,

My bad - we've started working on our project before the migration and I've overlooked both the fact that it happened and the message on the old repository. Will migrate soon and see whether it helps with our internal problems. Thank you for letting me know!

neurolabusc:
@pFornagiel since you are using Claude, when you move to the new API, you can tell Claude to use this [cheat sheet](https://github.com/niivue/mono/blob/fd703c92154d695676fdd038da69185005f09c4f/packages/niivue/docs/convert.md?plain=1#L4). For all of my projects, it ported from legacy to the new API in one shot.

Are you going to be in OHBM in Bordeaux this week? Happy to meet, alternatively, happy to meet via zoom sometime and describe the rationale and benefits for the refactored code.

pFornagiel:
@neurolabusc thanks for the headsup, managed to port already! Not planning to go to Bordeaux conference, as I do not have much expertise in this field unfortunately. In our project we've been using the library to perform visualisation of dental Nifti volumes from [ToothFairy3 dataset](https://ditto.ing.unimore.it/toothfairy3/) with pretty good results.

Some of the patches made in legacy version still apply to the `v1.0.0-rc.9`. Probably going to run some benchmarks and will see whether the changes are applicable - if those are sensible and not project-specific, I will be happy to continue discussion under the opened PR on the mono-repository.

PR #1617, opened 2026-08-07
Fix Zenodo DOI badge

CodyCBakerPhD:
It's a somewhat common issue to see the default Zenodo DOI badges (as they provide through their site) fail to render as shown below

image

this is a simple fix that maintains the appearance and renders more reliably

(sometimes the Zenodo badge will show up if you wait long enough, or refresh page enough times, or some other magical conditions - it's just not super stable)


CodyCBakerPhD:
Witness the fixed rendering on: https://github.com/CodyCBakerPhD/niivue/tree/patch-1

